PR #4, opened 2024-03-18
Relate to keystone to get credentials for exporter

samuelallan72:
Save these credentials to a clouds.yaml file,
so they can be used by openstack-exporter.

If keystone in your test environment uses https, you may need to manually set the `ssl_ca` config.

internal ticket: [BSENG-2172](https://warthogs.atlassian.net/browse/BSENG-2172)

[BSENG-2172]: https://warthogs.atlassian.net/browse/BSENG-2172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -176,7 +176,7 @@ class OpenstackExporterOperatorCharm(ops.CharmBase):
 
         logger.info("Keystone credentials are available, starting services.")
         self._write_cloud_config(data)
-        snap_service.start()
+        snap_service.restart_and_enable()
 
     def _on_install(self, _: ops.InstallEvent) -> None:
         """Handle install charm event."""
--- a/src/service.py
+++ b/src/service.py
@@ -20,8 +20,19 @@ class SnapService:
         """Return True if the snap service is active."""
         return self.snap_client.services.get(self.snap_service, {}).get("active", False)
 
-    def start(self) -> None:
-        """Start and enable the snap service."""
+    def restart_and_enable(self) -> None:
+        """Restart and enable the snap service.
+
+        Ensure:
+        - service is running
+        - service is enabled
+        - service is restarted if already running to apply updated config
+        """
+        # This is safe to always run,
+        # because restarting when service is disabled has no effect,
+        # and restarting when enabled but stopped has the same effect as start.
+        self.snap_client.restart([self.snap_service])
+        # This is idempotent, so ok to always run to ensure it's started and enabled
         self.snap_client.start([self.snap_service], enable=True)
 
     def stop(self) -> None:


src/charm.py
```diff
@@ -9,19 +9,132 @@
...
         """Initialize the charm."""
         super().__init__(*args)
+        self.framework.observe(self.on.upgrade_charm, self._on_upgrade)
+        self.framework.observe(self.on.config_changed, self._on_config_changed)
+        self.framework.observe(self.on.credentials_relation_changed, self._on_credentials_changed)
+        self.framework.observe(self.on.collect_unit_status, self._on_collect_unit_status)
+
+    def _on_upgrade(self, _: UpgradeCharmEvent) -> None:
+        """Reconfigure everything on upgrade."""
+        self._configure()
```
chanchiwai-ray, line 40: This is unnecessary, please remove it. Charm upgrade will run config event https://juju.is/docs/sdk/upgrade-charm-event

src/charm.py
```diff
@@ -9,19 +9,132 @@
...
+        https://docs.openstack.org/keystone/latest/contributor/http-api.html
+        """
+        auth_url = "{protocol}://{hostname}:{port}/v3".format(
+            protocol=data["service_protocol"],
+            hostname=data["service_hostname"],
+            port=data["service_port"],
+        )
+        contents = {
+            "clouds": {
+                CLOUD_NAME: {  # the name here must be passed to --cloud option of openstack-exporter
```
chanchiwai-ray, line 96: This is not necessary comment

chanchiwai-ray (changes requested):
Small issues and some questions

README.md
```diff
@@ -7,3 +7,21 @@ the Prometheus monitoring system. This operator works with the Canonical
 Observability Stack (COS).
 
 More information: https://charmhub.io/openstack-exporter-operator
+
+## Development
```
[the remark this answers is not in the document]
chanchiwai-ray: A link to `CONTRIBUTION.md` would be nice

src/charm.py
chanchiwai-ray: well, at least these will be hardcoded in the charm anyways

README.md
```diff
@@ -7,3 +7,5 @@ the Prometheus monitoring system. This operator works with the Canonical
 Observability Stack (COS).
 
 More information: https://charmhub.io/openstack-exporter-operator
+
+For developer documentation, please see [./CONTRIBUTING.md](./CONTRIBUTING.md).
```
jneo8, line 11: **nitpick (non-blocking):** 

Consider move this to another independent PR.

charmcraft.yaml
```diff
@@ -16,13 +16,17 @@ bases:
...
+  credentials:
+    interface: keystone-admin
+
+config:
+  options:
+    ssl_ca:
+      default:
+      type: string
+      description: |
+        Custom SSL CA for keystone if required.
```
chanchiwai-ray, line 29: Sorry, one final spot: we should probably document this a bit more (it's a base64 encoded string or plain multi-line string (because it's a CA)). Users coming from the old [charm](https://git.launchpad.net/charm-prometheus-openstack-exporter/tree/src/reactive/openstack_exporter.py#n192) will assume that it's base64 encoded. For me, multi-line string is kind of hard to use

src/charm.py
```diff
@@ -9,19 +9,127 @@
...
+                "service_username",
+                "service_password",
+                "service_project_name",
+                "service_project_domain_name",
+                "service_user_domain_name",
+                "service_region",
+            ]
+        )
+
+    def update_clouds_yaml(self, data: dict[str, str]) -> None:
```
jneo8, line 77: **nitpick (non-blocking):** 

The current function name does not clearly indicate its purpose. User need background knowledge that the exporter needs a 'clouds.yaml' config file. A more descriptive name like 'generate_exporter_config' might be better

src/charm.py
```diff
@@ -9,19 +9,127 @@
...
+        if data:
+            logger.info("Keystone credentials are available, starting services.")
+            self.update_clouds_yaml(data)
+            # TODO: when openstack-exporter service is added, restart it here
+        else:
+            logger.info("Keystone credentials are not available, stopping services.")
+            # TODO: when openstack-exporter service is added, stop it here
+
+    def _is_keystone_data_ready(self, data: dict[str, str]) -> bool:
+        """Check if all the data is available from keystone."""
```
jneo8, line 61: **nitpick:** 

You probably want to add doc-string here to explain the meaning of parameter `data`.

jneo8 (changes requested):

src/charm.py
```diff
@@ -9,19 +9,132 @@
...
+                    "auth": {
+                        "username": data["service_username"],
+                        "password": data["service_password"],
+                        "project_name": data["service_project_name"],
+                        "project_domain_name": data["service_project_domain_name"],
+                        "user_domain_name": data["service_user_domain_name"],
+                        "auth_url": auth_url,
+                    },
+                    "verify": data["service_protocol"] == "https",
+                    "cacert": self.config["ssl_ca"] or None,
```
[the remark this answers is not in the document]
jneo8: +1 good point from Ray.

Provide through the config is fine now. The behavior is as same as the *charm-prometheus-openstack-exporter*.

I think we can create a issue here as a backlog.

charmcraft.yaml
chanchiwai-ray: Sure, that's fine. Just want to point that out loudly

charmcraft.yaml
```diff
@@ -16,13 +16,20 @@ bases:
...
-# config:
-#   options:
-#     snap_channel:
-#       default: "latest/stable"
-#       type: string
-#       description: |
-#           If install_method is set to "snap" this option controls channel name.
+requires:
+  credentials:
+    interface: keystone-admin
```
chanchiwai-ray, line 21: question: what is the difference between keystone-admin and keystone-credential?

chanchiwai-ray: oh, side note: I was just wanting to leave a comment, but the status become request-change

src/charm.py
```diff
@@ -9,19 +9,133 @@
...
```
chanchiwai-ray, line 110: `self.config["ssl_ca"] or None` should be changed to `self.config.get("ssl_ca", "")` since the default value of that config is `unset` which is causes 

```
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-openstack-exporter-3/charm/./src/charm.py", line 240, in <module>
    ops.main(OpenstackExporterOperatorCharm)
  File "/var/lib/juju/agents/unit-openstack-exporter-3/charm/venv/ops/main.py", line 471, in __call__
    return main(charm_class, use_juju_for_storage=use_juju_for_storage)
  File "/var/lib/juju/agents/unit-openstack-exporter-3/charm/venv/ops/main.py", line 454, in main
    framework.reemit()
  File "/var/lib/juju/agents/unit-openstack-exporter-3/charm/venv/ops/framework.py", line 875, in reemit
    self._reemit()
  File "/var/lib/juju/agents/unit-openstack-exporter-3/charm/venv/ops/framework.py", line 955, in _reemit
    custom_handler(event)
  File "/var/lib/juju/agents/unit-openstack-exporter-3/charm/./src/charm.py", line 210, in _on_config_changed
    self.configure(event)
  File "/var/lib/juju/agents/unit-openstack-exporter-3/charm/./src/charm.py", line 160, in configure
    self.write_cloud_config(data)
  File "/var/lib/juju/agents/unit-openstack-exporter-3/charm/./src/charm.py", line 107, in write_cloud_config
    "cacert": self.config["ssl_ca"] or None,
  File "/var/lib/juju/agents/unit-openstack-exporter-3/charm/venv/ops/model.py", line 796, in __getitem__
    return self._data[key]
KeyError: 'ssl_ca'
```

chanchiwai-ray: oh, why not `self.config.get("ssl_ca")` without `or None`

chanchiwai-ray: I don't get it.. Isn't the one you describing the same as `self.config.get("ssl_ca")`?

If `ssl_ca` is unset, set to null, or empty string, `self.config.get("ssl_ca")` will return `None`

chanchiwai-ray: that also works, thanks!

src/charm.py
```diff
@@ -9,19 +9,133 @@
...
+        This is an idempotent method to configure everything.
+        """
+        data = self._get_keystone_data()
+
+        if data:
+            logger.info("Keystone credentials are available, starting services.")
+            self.write_cloud_config(data)
+            # TODO: when openstack-exporter service is added, restart it here
+        else:
+            logger.info("Keystone credentials are not available, stopping services.")
```
chanchiwai-ray, line 57: Do we need to defer here? I can imagine if we can't get the relation data the first try when credentials changed, then we need to remove-relation and relate again

chanchiwai-ray (dismissed):
But overall it looks good to me. I don't want to be picky right now and block the PR here

src/charm.py
chanchiwai-ray: Okay, thanks

charmcraft.yaml
```diff
@@ -16,13 +16,20 @@ bases:
...
+config:
+  options:
+    ssl_ca:
+      default:
+      type: string
+      description: |
+        Custom SSL CA for keystone if required.
+
+        The format should be the contents of a PEM encoded file.
+        (no base64 encoding required).
```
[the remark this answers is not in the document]
jneo8: Sure.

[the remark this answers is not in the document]
jneo8: #9

jneo8 (dismissed):

src/charm.py
```diff
@@ -35,8 +33,10 @@
 
 
 # Snap config options global constants
+# This is to match between openstack-exporter and the entry in clouds.yaml
 CLOUD_NAME = "openstack"
-OS_CLIENT_CONFIG = USER_HOME / "clouds.yaml"
+# store the clouds.yaml where it's easily accessible by the openstack-exporter snap
+OS_CLIENT_CONFIG = "/var/snap/golang-openstack-exporter/common/clouds.yaml"
```
chanchiwai-ray, line 39: Nice find, I almost forgot the `SNAP_COMMON` directory

src/charm.py
```diff
@@ -94,8 +166,14 @@ def configure(self, event: ops.HookEvent) -> None:
...
-        # keystone and grafana-agent.
+        data = self._get_keystone_data()
+        if not data:
+            logger.info("Keystone credentials are not available, stopping services.")
+            snap_service.stop()
+            return
+
+        logger.info("Keystone credentials are available, starting services.")
+        self._write_cloud_config(data)
         snap_service.start()
```
chanchiwai-ray, line 179: We might need a `snap_service.restart()`, I tested manually with the snap: starting an already started service will not take effect

chanchiwai-ray (approved):
Thanks, overall looks good, I left a comment about `snap_service.restart()`.

(optional and opinionated) Can we make the non-event function without the `_` prefix? It just looks nicer to me, and make me easier to visually distinguish the event callbacks and normal functions

src/charm.py
chanchiwai-ray: oh, it should have been `changes-requested` because it's kind of a bug

chanchiwai-ray:
> > Can we make the non-event function without the _ prefix? It just looks nicer to me, and make me easier to visually distinguish the event callbacks and normal functions
> 
> So ideally all the other methods should have the `_` prefix too, as this is used as a convention to show that they are internal methods and shouldn't be called from outside the class.

Yes, I understand that, so it's optional

chanchiwai-ray (approved):
Thanks, LGTM.

Did some basic testing with keystone:

- [x] installed with resource -> Blocked: missing keystone relation
- [x] relate to keystone ->  Active
- [x] curl the endpoint -> Blocked: snap service is not running (after update-status) (note this is expected with only keystone)
- [ ] changing config ssl_ca render the `clouds.yaml`

PR #11, opened 2024-03-20
Add integration with Grafana Agent

samuelallan72:
Add integration with Grafana Agent

- pull in the cos_agent interface library
- configure the relation with path and port for openstack-exporter metrics

## Testing

Some helpful commands for testing:

```
charmcraft -v pack
juju deploy ./openstack-exporter_ubuntu-22.04-amd64.charm --resource openstack-exporter=golang-openstack-exporter_1.6.0_amd64.snap
juju deploy grafana-agent
juju relate openstack-exporter grafana-agent
```

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -61,6 +61,12 @@ class OpenstackExporterOperatorCharm(ops.CharmBase):
         self.framework.observe(self.on.config_changed, self._on_config_changed)
         self.framework.observe(self.on.collect_unit_status, self._on_collect_unit_status)
         self.framework.observe(self.on.credentials_relation_changed, self._on_credentials_changed)
+        self.framework.observe(
+            self.on.cos_agent_relation_changed, self._on_cos_agent_relation_changed
+        )
+        self.framework.observe(
+            self.on.cos_agent_relation_broken, self._on_cos_agent_relation_broken
+        )
 
     def _is_keystone_data_ready(self, data: dict[str, str]) -> bool:
         """Check if all the data is available from keystone.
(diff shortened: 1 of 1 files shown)


src/charm.py
```diff
@@ -46,6 +47,13 @@ def __init__(self, *args: tuple[Any]) -> None:
         """Initialize the charm."""
         super().__init__(*args)
 
+        self._grafana_agent = COSAgentProvider(
```
chanchiwai-ray, line 50: It's worth thinking should we start the snap service without this relation. We generally think the exporter service should be started only when it's required (i.e. there is consumers for the metric). Can you add some logic in the `configure` method to properly handle that?

chanchiwai-ray (changes requested):

CONTRIBUTING.md
chanchiwai-ray, line 1: Thanks for the doc updates, can you move this to a separate PR? I need these to create workflow for automatic updating the libs

update_charm_libs.sh
chanchiwai-ray, line 1: Also move this to a new PR

chanchiwai-ray (changes requested):

chanchiwai-ray (dismissed):
LGTM, thanks

CONTRIBUTING.md
chanchiwai-ray: Thanks

jneo8 (approved):

chanchiwai-ray (approved):

jneo8 (approved):

PR #19, opened 2024-03-20
Document charm libs and provide command to update

samuelallan72:


chanchiwai-ray (approved):
LGTM

tox.ini
```diff
@@ -59,3 +59,9 @@ deps =
     mypy
 commands =
     mypy --install-types --non-interactive .
+
+[testenv:update-charm-lib]
```
jneo8, line 63: **issue (blocking):** 

The command indirect result is the creation of an additional Python environment within .tox solely for its execution. If storage for the command is needed, a Makefile would be a more appropriate choice

jneo8 (changes requested):

tox.ini
chanchiwai-ray: A script like previously will be fine (e.g. `scripts/update-charm-lib.sh`), using makefile for just a single target is overkill.

chanchiwai-ray: unless you have more targets in mind?

chanchiwai-ray: we won't need scripts for update dashboard files from openstack-exporter-k8s, it's very simple commands, can be in the workflow itself

chanchiwai-ray: that being said, I don't oppose plan for the future

chanchiwai-ray: Sure

chanchiwai-ray: well, we can make it independent though, we create `scripts/*.sh` and let makefile to call the scripts

jneo8: > well, we can make it independent though, we create scripts/*.sh and let makefile to call the scripts

This one or makefile are work for me.

chanchiwai-ray (approved):

jneo8 (approved):

chanchiwai-ray:
Rebasing with web UI create a `Unverified` commit and blocks the merge ¯\_(ツ)_/¯

Issue #21, opened 2024-03-21
when "credentials" relation does not exist the exporter should be stopped


chanchiwai-ray:
When the "credentials" relation is initially joined, the snap service starts. But when the "credentials" relation is removed, the snap service stays running even though the charm is in `Blocked:  Keystone is not related`

PR #23, opened 2024-03-25
Add callback for credentials relation broken

samuelallan72:
And refactor to merge duplicate code.
Several events only need to call the configure method, so the event handlers can be removed/merged.

Fixes: #21


chanchiwai-ray (approved):
Thanks, LGTM

jneo8 (approved):

PR #30, opened 2024-04-04
Fix missing nova metrics from openstack-exporter

samuelallan72:
Previously this error was seen in the logs:

  ts=2024-04-03T23:09:00.715Z caller=exporter.go:126 level=error err="Failed to collect metric for exporter" exporter=nova error="failed to collect metric: running_vms, error: CPUInfo has unexpected type: <nil>"

Causing `openstack_nova_running_vms` and a selection of other hypervisor related metrics to be missing from the output.

This applies the workaround noted in https://github.com/openstack-exporter/openstack-exporter/issues/268

### DIFF
--- a/src/service.py
+++ b/src/service.py
@@ -1,5 +1,6 @@
 """Utility module to help manage the snap service with guarding functions."""
 
+import os
 from logging import getLogger
 from typing import Any, Optional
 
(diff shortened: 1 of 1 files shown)


aieri (approved):
This looks ok, but I would much prefer to fix this as high up the chain as possible (so probably in the exporter) to maximise the impact on the community

chanchiwai-ray (approved):
LGTM, however I have a question: will it cause any incompatibility issue for other openstack releases (e.g. causing similar issue if you set it to 2.87)

PR #36, opened 2024-04-09
test: Init zaza integration

jneo8:
- refactor file structure to make zaza work properly
- Download snap file from github release assets


Makefile
```diff
@@ -1,11 +1,21 @@
...
 help:
 	@echo "This project supports the following targets"
 	@echo ""
 	@echo " make                            - show help text"
 	@echo " make update-charm-libs          - update charm's libraries"
 	@echo " make check-dashboard-updates    - check if there's a new dashboard from the upstream"
+	@echo " make clean - remove unneeded files"
+	@echo " make download-snap - download snap release from github release assets"
+	@echo " make build - build the charm"
+	@echo " make integration - run the tests defined in the integration subdirectory"
```
chanchiwai-ray, line 18: Can you align the text?

tests/integration/requirements.txt
```diff
@@ -1,4 +1,4 @@
 python-openstackclient
 
-# FIXME: when zaza support libjuju 3.x, remove the pin
 git+https://github.com/openstack-charmers/zaza.git@libjuju-3.1#egg=zaza
+juju<4
```
chanchiwai-ray, line 4: what is this ?

tox.ini
```diff
@@ -60,9 +61,9 @@ commands =
 
 [testenv:integration]
 description = Run integration tests
-changedir = {toxinidir}/tests/integration/
```
chanchiwai-ray, line 63: why do you remove this, and the following line? Is there any reason blocking the zaza test?

Those two lines are added to avoid file renames like you did in the up above (i.e making a unnecessary nested `tests/integration/tests` directory), and you can simply use `tests/integration`.

chanchiwai-ray (changes requested):
Small comments

Makefile
```diff
@@ -15,4 +25,23 @@ update-charm-libs:
...
+	@rm -rf ${PROJECTPATH}/*.snap
+
+build: clean
+	@echo "Building charm"
+	@charmcraft -v pack
+
+download-snap:
+	wget -q https://github.com/canonical/openstack-exporter-operator/releases/download/rev2/golang-openstack-exporter_amd64.snap -O ./golang-openstack-exporter_amd64.snap
+
+integration: download-snap build
```
chanchiwai-ray, line 44: the build -> clean dependency will remove the snap file

Makefile
```diff
@@ -15,4 +25,23 @@ update-charm-libs:
...
+	CHARM_LOCATION=${CHARM_LOCATION} CHARM_SNAP_LOCATION=${CHARM_SNAP_LOCATION} tox -e integration
```
chanchiwai-ray, line 45: Missing `charm_name` env  see `local-charm-overlay.yaml.j2`

tox.ini
jneo8: `tests/integration/tests/` nested directory is needed unfortunately. See https://github.com/openstack-charmers/zaza/blob/b22c2eed4c322f1dfc14ffb2d31e0dd18c911a40/zaza/charm_lifecycle/deploy.py#L38

tests/integration/requirements.txt
jneo8: It's the same, it will use juju 3.1.x

Makefile
jneo8: `charm_name` env is assign by zaza.

tests/integration/requirements.txt
```diff
@@ -1,4 +1,3 @@
 python-openstackclient
 
-# FIXME: when zaza support libjuju 3.x, remove the pin
```
chanchiwai-ray, line 3: Why remove this comment? I think it's still not fixed

chanchiwai-ray (changes requested):
One small question

tox.ini
chanchiwai-ray: Nice find!

chanchiwai-ray (dismissed):
LGTM

Makefile
```diff
@@ -15,4 +25,23 @@ update-charm-libs:
 check-dashboard-updates:
 	./scripts/check-dashboard-updates.sh
 
-.PHONY: help update-charm-libs check-dashboard-updates
+clean:
+	@echo "clean files"
+	@git clean -ffXd -e '!.idea' -e '!.vscode'
```
rgildein, line 30: Why we need this? I can have my own test yaml (e.g. bundles) in repo and like this it will be removed.

Makefile
```diff
@@ -15,4 +25,23 @@ update-charm-libs:
...
+	@charmcraft clean
+	@echo "Remove download snap"
+	@rm -rf ${PROJECTPATH}/*.snap
+
+build: clean
+	@echo "Building charm"
+	@charmcraft -v pack
+
+download-snap:
+	wget -q https://github.com/canonical/openstack-exporter-operator/releases/download/rev2/golang-openstack-exporter_amd64.snap -O ./golang-openstack-exporter_amd64.snap
```
rgildein, line 42: Can we use snap directly here, e.g. `snap download golang-openstack-exporter --channel stable`.

Makefile
```diff
@@ -15,4 +25,23 @@ update-charm-libs:
...
 
-.PHONY: help update-charm-libs check-dashboard-updates
+clean:
+	@echo "clean files"
+	@git clean -ffXd -e '!.idea' -e '!.vscode'
+	@echo "Cleaning existing build"
+	@rm -rf ${PROJECTPATH}/${CHARM_NAME}*.charm
+	@charmcraft clean
+	@echo "Remove download snap"
+	@rm -rf ${PROJECTPATH}/*.snap
```
rgildein, line 35: I think we should follow same approach as we are doing with charms, e.g. removing `${PROJECTPATH}/golang-openstack-exporter*.snap` instead of any snap.

rgildein (changes requested):

Makefile
jneo8: https://github.com/search?q=org%3Acanonical%20%40git%20clean%20-ffXd%20-e%20%27!.idea%27%20-e%20%27!.vscode%27&type=code

Does not this our practice to clean the files?

Makefile
jneo8: This project should not include any snap, and the snap can be download from git release assert easily.

Makefile
rgildein: I see, maybe we should not do it. What do you say?

Makefile
rgildein: Ack. Maybe we can add some note or even better raise an issue, that we should use upstream if it will be released.

Makefile
rgildein: It's up to me, what I put in the repo directory :D, so that's why I think it would be great not remove any of my files.

jneo8: Ok, let me update it.

rgildein (approved):
LGTM, thanks

chanchiwai-ray (approved):

Makefile
rgildein: @jneo8 Can you open an issue for this. Thanks

PR #39, opened 2024-04-11
Add new script to sync dashboards

samuelallan72:
With a flag to check if dashboards are updated
(so `make check-dashboard-updates` still works).

Also run the script to sync the dashboards.

### DIFF
--- a/scripts/sync-dashboards
+++ b/scripts/sync-dashboards
@@ -5,7 +5,8 @@ import sys
 import argparse
 from urllib.request import urlretrieve
 
-# sync the dashboards from https://opendev.org/openstack/sunbeam-charms/src/branch/main/charms/openstack-exporter-k8s/src/grafana_dashboards
+# Sync the dashboards from https://opendev.org/openstack/sunbeam-charms/src/branch/main/charms/openstack-exporter-k8s/src/grafana_dashboards
+# Note: the base_url below is expected to be a 404 by itself.
 BASE_URL = "https://opendev.org/openstack/sunbeam-charms/raw/branch/main/charms/openstack-exporter-k8s/src/grafana_dashboards/"
 BASE_LOCAL_PATH = "./src/grafana_dashboards/"
 DASHBOARDS = [


scripts/sync-dashboards
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env python3
+
+import os
+import sys
+import argparse
+from urllib.request import urlretrieve
+
+# sync the dashboards from https://opendev.org/openstack/sunbeam-charms/src/branch/main/charms/openstack-exporter-k8s/src/grafana_dashboards
+BASE_URL = "https://opendev.org/openstack/sunbeam-charms/raw/branch/main/charms/openstack-exporter-k8s/src/grafana_dashboards/"
```
chanchiwai-ray, line 9: Can you join the `BASE_URL` and `DASHBOARDS` as `DASHBOARS_URLS` (or some other name), the `BASE_URL`  on its own is a 404

scripts/sync-dashboards
```diff
@@ -0,0 +1,72 @@
...
+def sync_dashboard(url: str, local_path: str, check_mode: bool) -> bool:
+    """Sync dashboard from url.
+
+    If check_mode is True, don't actually update the file.
+    Return True if the dashboard changed (or would be changed).
+    """
+    new_dashboard_path, _ = urlretrieve(url)
+
+    changed = False
+    if os.path.exists(local_path):
```
chanchiwai-ray, line 27: I think we should not check the existence of `local_path`? since the remote is the source of truth. I want to confirm, what is the behavior you are expecting in the following case:

1. if there are new or missing files from "REMOTE", should we report difference?

scripts/sync-dashboards
```diff
@@ -0,0 +1,72 @@
...
+        with open(new_dashboard_path) as f1, open(local_path) as f2:
+            changed = f1.read() != f2.read()
+    else:
+        changed = True
+
+    if changed and not check_mode:
```
chanchiwai-ray, line 33: I would probably refactor the "sync" into a small functional call to avoid overloading a single function with two responsibility, but non-blocking since it's a small and simple script

chanchiwai-ray (changes requested):
Thanks for the refactoring, overall LGTM, but I have some comments and questions

jneo8 (changes requested):
I want to bring up different solution here.
Can't we just use `wget` and `diff` commands? I feel it will be more easy that using python.

chanchiwai-ray:
Spotted small issue: running the script in the script directory will not work properly

scripts/sync-dashboards
chanchiwai-ray: Yeah, they will not happen often or even won't happen, but still nice to do it if it's not difficult though (I don't want to over-complicate this script), otherwise it's not "sync-dashboards". I was more expecting `rsycn ... --delete <remote>  <src>` like behaviour

scripts/sync-dashboards
chanchiwai-ray: Thanks

jneo8:
> I want to bring up different solution here.
> Can't we just use `wget` and `diff` commands? I feel it will be more easier than using python.

Since it's a simple script and the scrips size will be similar. It's ok to go with python.

jneo8 (approved):

scripts/sync-dashboards
chanchiwai-ray: 👌🏼

scripts/sync-dashboards
chanchiwai-ray: 👌🏼

chanchiwai-ray (approved):
Overall LGTM, no blocking issues

PR #42, opened 2024-04-12
Uncomment links metadata and link to new docs

samuelallan72:
New docs are being developed at https://discourse.charmhub.io/t/openstack-exporter-docs-index/13876 - currently they're minimal / not much different to what's already in the description in metadata.

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -47,8 +47,6 @@ links:
   - https://github.com/canonical/openstack-exporter-operator/issues
   source:
   - https://github.com/canonical/openstack-exporter-operator
-  website:
-  - https://charmhub.io/openstack-exporter
 
 resources:
   openstack-exporter:


charmcraft.yaml
```diff
@@ -41,19 +41,14 @@ config:
...
-#   - https://github.com/canonical/openstack-exporter-operator
-#   website:
-#   - https://charmhub.io/openstack-exporter
+links:
+  documentation: https://discourse.charmhub.io/t/openstack-exporter-docs-index/13876
+  issues:
+  - https://github.com/canonical/openstack-exporter-operator/issues
+  source:
+  - https://github.com/canonical/openstack-exporter-operator
+  website:
```
chanchiwai-ray, line 50: It feels like we don't understand the "website" correctly: https://juju.is/docs/sdk/charmcraft-yaml#heading--links

> Link to charm project website outside of Charmhub, if available.

we should remove it because we don't have project website outside charmhub, or do we ?

chanchiwai-ray (changes requested):
Just read the doc about `links`, and have some comments

chanchiwai-ray (approved):

Pjack (approved):

PR #48, opened 2024-04-19
feat: Provide cache configuration

jneo8:
* Add support for snap `cache` and `cache-ttl` configurations.
* Change unit test basic class to pytest and use pytest-mocker

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -33,13 +33,23 @@ config:
         The openstack-exporter service (metrics endpoint for prometheus scraping)
         will listen at this port.
     ssl_ca:
-      default: ''
+      default: ""
       type: string
       description: |
         Custom SSL CA for keystone if required.
 
         The format should be the raw contents of a PEM encoded file.
         (no base64 encoding).
+    cache_ttl:
+      default: "300s"
+      type: string
+      description: |
+        TTL duration for cache expiry(eg. 10s, 11m, 1h)
+    cache:
+      default: false
+      type: boolean
+      description: |
+        Enable exporter cache globally
 
 
 links:
--- a/tests/unit/requirements.txt
+++ b/tests/unit/requirements.txt
@@ -1,3 +1,5 @@
 # For cos_agent library
 cosl
 pydantic < 2
+pytest
+pytest-mock
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -3,18 +3,55 @@
 #
 # Learn more about testing at: https://juju.is/docs/sdk/testing
 
-import unittest
-
 import ops
 import ops.testing
-from charm import OpenstackExporterOperatorCharm
+import pytest
+from charm import CLOUD_NAME, OS_CLIENT_CONFIG, SNAP_NAME, OpenstackExporterOperatorCharm
 
 
-class TestCharm(unittest.TestCase):
-    def setUp(self):
+class TestCharm:
+    def setup_method(self, _):
         self.harness = ops.testing.Harness(OpenstackExporterOperatorCharm)
-        self.addCleanup(self.harness.cleanup)
+
+    def teardown_method(self, _):
+        self.harness.cleanup()
+
+    @pytest.mark.parametrize(
+        "config",
+        [
+            ({"cache": False, "cache_ttl": "100s"}),
+            ({"cache": True, "cache_ttl": "200s"}),
+        ],
+    )
+    def test_config_changed(self, config, mocker):
+        mock_get_installed_snap_service = mocker.patch("charm.get_installed_snap_service")
+        mock_snap_service = mocker.Mock()
+        mock_get_installed_snap_service.return_value = mock_snap_service
+
         self.harness.begin()
 
-    def test_null(self):
-        """Remove me."""
+        mock_get_keystone_data = mocker.Mock()
+        mock_keystone_data = mocker.Mock()
+        mock_get_keystone_data.return_value = mock_keystone_data
+        self.harness.charm._get_keystone_data = mock_get_keystone_data
+
+        mock_write_cloud_config = mocker.Mock()
+        self.harness.charm._write_cloud_config = mock_write_cloud_config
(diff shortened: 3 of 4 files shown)


tests/unit/requirements.txt
```diff
@@ -1,3 +1,5 @@
 # For cos_agent library
 cosl
 pydantic < 2
+pytest
+pytest-mock
```
rgildein, line 5: Can we use mocking from unittests? We are not using any feature of this library, so I would prefer to avoid adding it as decency.

tests/unit/test_charm.py
```diff
@@ -3,18 +3,55 @@
...
+    @pytest.mark.parametrize(
+        "config",
+        [
+            ({"cache": False, "cache_ttl": "100s"}),
+            ({"cache": True, "cache_ttl": "200s"}),
+        ],
+    )
+    def test_config_changed(self, config, mocker):
+        mock_get_installed_snap_service = mocker.patch("charm.get_installed_snap_service")
+        mock_snap_service = mocker.Mock()
```
rgildein, line 28: Can we use spec_set to define what we are mocking, at-least at some of this mock object.

tests/unit/requirements.txt
jneo8: I tend to not mix the unittest and pytest. Let's keep it this style?

charmcraft.yaml
```diff
@@ -33,13 +33,23 @@ config:
...
       description: |
         Custom SSL CA for keystone if required.
 
         The format should be the raw contents of a PEM encoded file.
         (no base64 encoding).
+    cache_ttl:
+      default: "300s"
+      type: string
+      description: |
+        TTL duration for cache expiry(eg. 10s, 11m, 1h)
```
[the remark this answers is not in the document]
jneo8: Updated

tests/unit/test_charm.py
jneo8: Updated

tests/unit/requirements.txt
rgildein: What mixing, importing patch and Mock, MagicMock directly is not mixing. This is definitely not blocking comment.

tests/unit/test_charm.py
```diff
@@ -3,18 +3,55 @@
...
+    @pytest.mark.parametrize(
+        "config",
+        [
+            ({"cache": False, "cache_ttl": "100s"}),
+            ({"cache": True, "cache_ttl": "200s"}),
+        ],
+    )
+    def test_config_changed(self, config, mocker):
+        mock_get_installed_snap_service = mocker.patch("charm.get_installed_snap_service")
+        mock_snap_service = mocker.Mock(spec_set=["configure", "restart_and_enable", "stop"])
```
rgildein, line 28: This should be `spec_set=SnapService`, there is no reason defining methods manually.

tests/unit/test_charm.py
```diff
@@ -3,18 +3,55 @@
...
+        mock_get_installed_snap_service.return_value = mock_snap_service
+
         self.harness.begin()
 
-    def test_null(self):
-        """Remove me."""
+        mock_get_keystone_data = mocker.Mock()
+        mock_keystone_data = mocker.Mock()
+        mock_get_keystone_data.return_value = mock_keystone_data
+        self.harness.charm._get_keystone_data = mock_get_keystone_data
```
rgildein, line 36: Checking this function in [charm.py](https://github.com/canonical/openstack-exporter-operator/blob/main/src/charm.py#L125) where it returns dictionary, so why we are using mock object here?

rgildein: I think that lines 33-36 should be changed to simply `self.harness.charm._get_keystone_data.return_value =  exp_keystone_data` and then use in line 55 `mock_write_cloud_config.assert_called_with(exp_keystone_data)`..

tests/unit/test_charm.py
```diff
@@ -3,18 +3,55 @@
...
+
+        mock_write_cloud_config = mocker.Mock()
+        self.harness.charm._write_cloud_config = mock_write_cloud_config
```
rgildein, line 39: Why not using `self.harness.charm._write_cloud_config.assert_called_with(...)`, why we need to create mock object and call `self.harness.charm._write_cloud_config = ...`? Should we rather properly mock this object and not change it behavior with assigning something else, using `obj = ...` instead of `path.object(self.harness.charm, "_write_cloud_config")`?

jneo8: I think we don't care the details of `_write_cloud_config` in this unit test case. We only make sure it's been called with right arguments.

tests/unit/test_charm.py
jneo8: Updated

tests/unit/test_charm.py
jneo8: Updated

tests/unit/requirements.txt
jneo8: Since it's not a blocking comment, I would like to resolve this one and keep what we have right now.

tests/unit/test_charm.py
rgildein: I was thinking about something like this:
```python
exp_keystone_data = {"key": "value"}
self.harness.charm._get_keystone_data.return_value =  exp_keystone_data
...
mock_write_cloud_config.assert_called_with(exp_keystone_data)
```

jneo8: Doesn't it the same? We only verify it's used as argument for `write_cloud_config`.

rgildein: I would be more understandable what we are doing, but yes you can pass eny object there even None.

jneo8: Sorry I don't think it's necessary. The Mock object has unique id, it's enough to verify what we want to test.

rgildein: I do not think this is good approach to mock thinks, but I'll not block this PR.

chanchiwai-ray (approved):

Pjack (approved):

PR #51, opened 2024-04-23
Support Juju 3.4

chanchiwai-ray:
- Changing to use of Juju 3.4 controller for testing instead of Juju 3.1
- Add workflow concurrency to avoid running multiple workflows on a single PR.

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -16,9 +16,8 @@ jobs:
     uses: ./.github/workflows/_quality_check.yaml
     secrets: inherit
     with:
-      tox-version: ""
+      tox-version: ""  # latest tox
       python-version: "3.10"
-      juju-channel: "3.4/stable"
 
   check-file-updates:
     name: Check updates for charm files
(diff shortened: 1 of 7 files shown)


rgildein (approved):
LGTM

rgildein (changes requested):
Now, I realized that you did not changed usage of zaza from master. Can you do so, thanks.

.github/workflows/pull-request.yaml
```diff
@@ -5,16 +5,19 @@ on:
...
+
 jobs:
 
   check-code-quality:
     name: Run charm code quality checks
     uses: ./.github/workflows/_quality_check.yaml
     secrets: inherit
     with:
-      tox-version: ""
+      tox-version: ""  # latest tox
```
rgildein, line 19: not-blocking: the default value should be changed in the bootstack-actions, and not here.

aieri (approved):

.github/workflows/pull-request.yaml
chanchiwai-ray: Yeah, we need to update `func.yaml@v3`. Resolving this for now, will open another PR later when `func.yaml@v3` is fixed

rgildein: It was fixed in the all version of bootstack-actions.

PR #54, opened 2024-05-01
Sync dashboards

### LINKED ISSUE #47: `Free (resources) per host` is too easy to turn into red
It looks like there are some logic of changing the color of the panel based on the data. For example with the `Free vCPUs per host` panel, it will turn into red when free vCPU per host is less than 100.

It's a little bit too aggressive to mark it as red (critical) to me. It might be better to use percentages instead of absolute values.

https://github.com/canonical/openstack-exporter-operator/blob/882043f084a4b9164e944a964e366e12cd551e22/src/grafana_dashboards/overview-openstack-exporter-dashboard.json#L1572-L1655

samuelallan72:
Includes patches:

- 917111: Remove absolute thresholds from overview dashboard | https://review.opendev.org/c/openstack/sunbeam-charms/+/917111

Fixes #47

### DIFF
--- a/src/grafana_dashboards/overview-openstack-exporter-dashboard.json
+++ b/src/grafana_dashboards/overview-openstack-exporter-dashboard.json
@@ -1035,10 +1035,6 @@
               {
                 "color": "green",
                 "value": null
-              },
-              {
-                "color": "red",
-                "value": 80
               }
             ]
           },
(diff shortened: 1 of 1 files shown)


rgildein (approved):
LGTM

chanchiwai-ray (approved):

PR #58, opened 2024-05-03
Add new release and promote workflow

rgildein:
Using official actions directly.


rgildein:
Cannot use fork here.

PR #61, opened 2024-05-06
Fix trigger for pull-request yaml

rgildein:


rgildein:
Merging without CI, since it's really no needed.

PR #66, opened 2024-05-28
chore: apply new github bot for jira and gh sync and remove the old jira_gh sync action

chanchiwai-ray:
Apply new GitHub bot for jira and gh sync, and optionally (if there's any) remove the old jira to gh sync action.


Pjack (approved):

chanchiwai-ray:
Func test keeps timing out

Pjack:
agree to skip the status check.. this MR is not related to the functional test at all.

PR #71, opened 2024-06-24
Run functional tests on LXD VMs

aieri:
- **Run functional tests on LXD VMs**
- **testing juju-crashdump**


aieri:
closing since we will wait for the private endpoint integration in the IS-hosted runners

PR #73, opened 2024-07-11
Add workflow to sync with Sunbeam

zxhdaze:
With the future implementation to integrate alert rules from sunbeam. We need to expand a bit the scope of the current approach.
This is just a draft PR to demo the new workflow sync files from sunbeam.
A sample PR created by this workflow can be found [here](https://github.com/canonical/openstack-exporter-operator/pull/75)

Currently, the check in this PR is failing because it is running from a forked repository. This workflow has been tested successfully in a branch at upstream: https://github.com/canonical/openstack-exporter-operator/actions/runs/10273103308/job/28426834216


.github/workflows/update_file.yaml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Checkout local repository
+        uses: actions/checkout@v4
+
+      - name: set git config
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          git config --global user.email "${GITHUB_ACTOR_ID}+${GITHUB_ACTOR}@users.noreply.github.com"
+          git config --global user.name "$(gh api /users/${GITHUB_ACTOR} | jq .name -r)"
+          git config -l
```
zxhdaze, line 22: I set this for my forked repo, I wonder if we need to do the same thing for the original repo too.

.github/workflows/update_file.yaml
```diff
@@ -0,0 +1,38 @@
...
+          
+      - name: Check pull from sunbeam
+        run: make sync-dashboards-alert-rules
+
+      - name: Create a PR for local changes
+        uses: peter-evans/create-pull-request@v6
+        id: cpr
+        with:
+          commit-message: "chore: update dashboards and alert rules"
+          title: "chore: update dashboards and alert rules"
```
zxhdaze, line 32: We are missing credentials for committer here.

.github/workflows/update_file.yaml
```diff
@@ -0,0 +1,38 @@
...
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          git config --global user.email "${GITHUB_ACTOR_ID}+${GITHUB_ACTOR}@users.noreply.github.com"
+          git config --global user.name "$(gh api /users/${GITHUB_ACTOR} | jq .name -r)"
+          git config -l
+          
+      - name: Check pull from sunbeam
+        run: make sync-dashboards-alert-rules
+
+      - name: Create a PR for local changes
```
samuelallan72, line 27: Does this handle the case where there are no new changes?  If there are no changes, it should not create the PR.

scripts/sync-dashboards-and-alert-rules.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
```
samuelallan72, line 1: This script need to be made executable and committed (git tracks executable bit for files):

```
chmod +x scripts/sync-dashboards-and-alert-rules.sh
git add scripts/sync-dashboards-and-alert-rules.sh
```

scripts/sync-dashboards-and-alert-rules.sh
```diff
@@ -0,0 +1,11 @@
...
+# Sync the dashboards  and alert rules from https://opendev.org/openstack/sunbeam-charms/src/branch/main/charms/openstack-exporter-k8s/src
+
+git remote add -f sunbeam https://opendev.org/openstack/sunbeam-charms.git
+git checkout -b staging-branch sunbeam/main
+git subtree split -P charms/openstack-exporter-k8s/src -b k8s-exporter
+git checkout k8s-exporter
+git rm -r $(ls | grep -vE '^grafana_dashboards$|^prometheus_alert_rules$')
+git commit -m "Remove unwanted files"
+git checkout main
+git subtree add -P src/sunbeam k8s-exporter --squash
```
samuelallan72, line 11: I'm not sure what the benefit of using a git subtree is here?  Could you explain why we're using git subtree instead of something like checking out the sunbeam-charms repo and copying the relevant files?  Unless there's a compelling reason, I don't like this method - it does weird things with the git history and it's hard to follow what's happening.  It's also a bit of a foot-gun having this work on the current checkout and main branch; running this accidentally could cause confusion.

Have we considered a workflow more like this (untested example only).  The local script and makefie target would do this:

```
git clone --depth 1 https://opendev.org/openstack/sunbeam-charms.git $tmpdir
rsync -av $tmpdir/charms/openstack-exporter-k8s/src/grafana_dashboards/ ./grafana_dashboards
# same rsync for alert_rules
```

Then the github workflow could:

- run that script
- if there are changes:
  - add and commit the change on a feature branch
  - open a PR with that new branch

If we wanted something fancy in the commit message we could link to upstream commits (or include the commit messages) for all commits that touched the dashboard/alert files since the last time the files were updated.

samuelallan72 (changes requested):

.github/workflows/update_file.yaml
```diff
@@ -0,0 +1,38 @@
...
+        id: cpr
+        with:
+          commit-message: "chore: update dashboards and alert rules"
+          title: "chore: update dashboards and alert rules"
+          body: |
+            Automated action to fetch the latest grafana dashboards and alert rules from Sunbeam. The branch of this PR 
+            will be overwritten during the next check. Unless you really know what you're doing, you 
+            most likely don't want to push any commits to this branch.
+          branch: chore/auto-libs
+          delete-branch: true
```
samuelallan72, line 38: Please configure your editor to auto-include the final trailing newline in files.  It's a good convention to follow, and most editors should do it by default.

samuelallan72 (changes requested):

scripts/update-charm-libs.sh
```diff
@@ -1,10 +0,0 @@
-#!/bin/sh
```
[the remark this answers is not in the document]
chanchiwai-ray: @zxhdaze for you information, we use this script to update the charm libs: https://github.com/canonical/openstack-exporter-operator/blob/main/CONTRIBUTING.md#fetching-libraries

.github/workflows/update_file.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: set git config
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          git config --global user.email "${GITHUB_ACTOR_ID}+${GITHUB_ACTOR}@users.noreply.github.com"
+          git config --global user.name "$(gh api /users/${GITHUB_ACTOR} | jq .name -r)"
+          git config -l
+          
+      - name: Check pull from sunbeam
```
chanchiwai-ray, line 24: I think this name should be clearer: may be something like `Pull dashboard and alert rules from sunbeam`?

chanchiwai-ray (commented):
For me, it's missing some context why we are changing to use `git subtree`, to check and update the dashboard files. Are you trying to address something the workflow can't do?

The existing workflow already includes the following:
- check dashboard files (can easily extend to check alert rules as well)
- sync dashboard files (already implemented) 
- periodically sync dashboards and alert rules (not implemented because it was considered not that useful at the time; dashboards usually doesn't change often when it's stable) (but can be easily implemented, just like you do here)

scripts/update-charm-libs.sh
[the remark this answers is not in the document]
zxhdaze: Ah yes it is deleted by accident.

scripts/sync-dashboards-and-alert-rules.sh
zxhdaze: We are considering using git subtree because we believe this approach might be more stable and bug-free when using git commands. And since we use --squash, we can keep the git history clean and customize the commit message.
This is just an exploration. We can definitely consider the workflow you suggested or stick with our current one if the majority prefers that.

.github/workflows/update_file.yaml
jneo8: I have same question here, there since no compare logic exists.

.github/workflows/update_file.yaml
```diff
@@ -0,0 +1,38 @@
+name: Update files
```
jneo8, line 1: **suggestion:** 

It will be nice if you can explain the whole workflow purpose here.

jneo8 (changes requested):
Some questions and suggestions.

.github/workflows/update_file.yaml
```diff
@@ -0,0 +1,38 @@
...
+        uses: peter-evans/create-pull-request@v6
+        id: cpr
+        with:
+          commit-message: "chore: update dashboards and alert rules"
+          title: "chore: update dashboards and alert rules"
+          body: |
+            Automated action to fetch the latest grafana dashboards and alert rules from Sunbeam. The branch of this PR 
+            will be overwritten during the next check. Unless you really know what you're doing, you 
+            most likely don't want to push any commits to this branch.
+          branch: chore/auto-libs
```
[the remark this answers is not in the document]
rgildein: Since branch will be automatically deleted, I do not see how it will be easier to track history?

.github/workflows/update_file.yaml
rgildein, line 22: I think we should use github application, because like this we will not have signed commit.

.github/workflows/update_file.yaml
```diff
@@ -0,0 +1,38 @@
...
+      - name: set git config
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          git config --global user.email "${GITHUB_ACTOR_ID}+${GITHUB_ACTOR}@users.noreply.github.com"
+          git config --global user.name "$(gh api /users/${GITHUB_ACTOR} | jq .name -r)"
+          git config -l
+          
+      - name: Check pull from sunbeam
+        run: make sync-dashboards-alert-rules
```
rgildein, line 25: I do not see reason for using Makefile and shell script here, like this we are adding more places which potentially need to be changed. To we should use `actions/checkout@v4` to get sunbeam repo (we can use the `sparse-checkout` option to get only subdirectory) and simply copy files we need.

If I understand this workflow properly, we do not need to have any check if new file was created. If git diff is empty, no PR will be created. Is that right?

scripts/sync-dashboards-and-alert-rules.sh
rgildein: I agree with @samuelallan72 here, I do not see reason why we're adding sunbeam as remote and then using subtree. Simply copy would be more appropriate. 

However, I think that script should be removed, and we can use GitHub actions directly.

rgildein (commented):
I'm providing only comment, since I do not want to block this.

I have two mayor comments:
- I do not like usage of any make command which is calling the shell script, since to understand what CI doing, I will need to check workflow -> Makefile -> scrip. I think that this problem is simple enough and we can use checkout action.
- We need to create PR with bot (GitHub application) and not with personal credential.

scripts/sync-dashboards-and-alert-rules.sh
samuelallan72: I don't think we should use github actions directly for this - a separate script would be helpful for developers to run locally on their machines for testing.  The script should do all the file syncing logic.  It shouldn't make any commits or mess with git; just copy files from the sunbeam repo.

rgildein: We can have helper script for developer, but in CI I think we should avoid using it. If we go with script, I would say that shell script with git command + cp / rsync would be better option than python code.

.github/workflows/update_file.yaml
Pjack: I have the same question. @zxhdaze could you help to answer it here? It seems we need to use github application which may increase the maintainence effort. I would like to understand what's the major benefit and what's the main issue we have now. Thanks!

scripts/update-charm-libs.sh
chanchiwai-ray, line 1: I think it's better to keep the "execute" permission

.github/workflows/update_file.yaml
chanchiwai-ray: I've tried https://github.com/actions/create-github-app-token, and it seems working fine.
Example: https://github.com/canonical/snap-tempest-automation/blob/main/.github/workflows/update-snapcraft.yaml#L77C13-L97

.github/workflows/update_file.yaml
```diff
@@ -0,0 +1,47 @@
...
+          
+      - name: Pull dashboard and alert rules from sunbeam
+        run: |
+          git clone --sparse https://opendev.org/openstack/sunbeam-charms.git
+          cd sunbeam-charms
+          git sparse-checkout set charms/openstack-exporter-k8s/src/grafana_dashboards charms/openstack-exporter-k8s/src/prometheus_alert_rules
+          cp -r charms/openstack-exporter-k8s/src/grafana_dashboards ../src/
+          cp -r charms/openstack-exporter-k8s/src/prometheus_alert_rules ../src/
+          cd ..
+          rm -rf sunbeam-charms
```
samuelallan72, line 34: I would like to see this in a separate script, so we can run it locally easily for testing. (eg. in the scripts directory)

.github/workflows/update_file.yaml
```diff
@@ -0,0 +1,47 @@
...
+        run: |
+          git config --global user.email "${GITHUB_ACTOR_ID}+${GITHUB_ACTOR}@users.noreply.github.com"
+          git config --global user.name "$(gh api /users/${GITHUB_ACTOR} | jq .name -r)"
+          git config -l
+          
+      - name: Pull dashboard and alert rules from sunbeam
+        run: |
+          git clone --sparse https://opendev.org/openstack/sunbeam-charms.git
+          cd sunbeam-charms
+          git sparse-checkout set charms/openstack-exporter-k8s/src/grafana_dashboards charms/openstack-exporter-k8s/src/prometheus_alert_rules
```
samuelallan72, line 30: Why do we have sparse-checkout here?  We're selectively copying below anyway, so sparse-checkout is not needed. :)

scripts/update-charm-libs.sh
samuelallan72: +1 please revert changes to this file

.github/workflows/update_file.yaml
```diff
@@ -0,0 +1,47 @@
...
+      - name: Create a PR for local changes
+        uses: peter-evans/create-pull-request@v6
+        id: cpr
+        with:
+          commit-message: "chore: update dashboards and alert rules"
+          title: "chore: update dashboards and alert rules"
+          body: |
+            Automated action to fetch the latest grafana dashboards and alert rules from Sunbeam. The branch of this PR 
+            will be overwritten during the next check. Unless you really know what you're doing, you 
+            most likely don't want to push any commits to this branch.
```
samuelallan72, line 45: To assist reviewers, it would be helpful to add links to the relevant directories on https://opendev.org/openstack/sunbeam-charms for reference.

samuelallan72 (changes requested):

Pjack:
> For me, it's missing some context why we are changing to use `git subtree`, to check and update the dashboard files. Are you trying to address something the workflow can't do?
> 
> The existing workflow already includes the following:
> 
> * check dashboard files (can easily extend to check alert rules as well)
> * sync dashboard files (already implemented)
> * periodically sync dashboards and alert rules (not implemented because it was considered not that useful at the time; dashboards usually doesn't change often when it's stable) (but can be easily implemented, just like you do here)

I have the same question. @zxhdaze could you help to answer it here? It seems we need to use github application which may increase the maintainence effort. I would like to understand what's the major benefit and what's the main issue we have now. Thanks!

jneo8 (changes requested):
Can we keep the sync dashboard script? Make both local sync & PR sync works may be more nice for development.

Also the explanation how this workflow works is not enough. Can you please add more description to either workflow file or document? The description should includes purpose of the workflow and the permission required to create the PR by action.

Also is running cron the best way? Can't we watch on the new commeit on sunbeam project?

.github/workflows/update_file.yaml
gabrielcocenza: If I understood correctly this fetch only two files instead of fetching all the sumbeam-charms project.

I'm in favor of  keeping this way

samuelallan72: Ah right, I missed `--sparse` clone above.

.github/workflows/pull-request.yaml
```diff
@@ -21,6 +21,7 @@ jobs:
       tox-version: ""
       python-version: "3.10"
 
-  check-file-updates:
+  update-file:
     name: Check updates for charm files
-    uses: ./.github/workflows/_check_file_update.yaml
+    uses: ./.github/workflows/update_file.yaml
```
aieri, line 26: suggestion: we are updating multiple files so this could be named more appropriately. For example "sync-from-sunbeam"

.github/workflows/update_file.yaml
```diff
@@ -0,0 +1,55 @@
...
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          git config --global user.email "${GITHUB_ACTOR_ID}+${GITHUB_ACTOR}@users.noreply.github.com"
+          git config --global user.name "$(gh api /users/${GITHUB_ACTOR} | jq .name -r)"
+          git config -l
+          
+      - name: Pull dashboard and alert rules from sunbeam
+        run: |
+          chmod +x scripts/pull_dashboard_and_alert_rules.sh
+          ./scripts/pull_dashboard_and_alert_rules.sh
```
aieri, line 26:
```suggestion
          bash scripts/pull_dashboard_and_alert_rules.sh
```
if you can't rely on the script being executable, just pass it to the interpreter

.github/workflows/update_file.yaml
```diff
@@ -0,0 +1,55 @@
...
+          app-id: ${{ secrets.APP_ID }}
+          owner: ${{ env.REPO_OWNER }}
+          private-key: ${{ secrets.APP_PEM }}
+          repositories: ${{ env.REPO_PATH }}
+      - name: Create pull request
+        uses: canonical/create-pull-request@main
+        with:
+          upsert: true  # Allow update PRs
+          github-token: ${{ steps.app_token.outputs.token }}
+          branch-name: chore/auto-libs  # create new branch with the changes as the source branch for the PR
```
aieri, line 40:
```suggestion
          branch-name: chore/sunbeam-sync  # create new branch with the changes as the source branch for the PR
```
iiuc we are not syncing any library

.github/workflows/update_file.yaml
```diff
@@ -0,0 +1,55 @@
...
+          repository: ${{ env.REPO_OWNER }}/${{ env.REPO_PATH }}
+          title: "Update dashboards and alerts"
+          body: "This pull request was generated by https://github.com/canonical/snap-tempest-automation"
```
aieri, line 43: needs updating

.github/workflows/update_file.yaml
```diff
@@ -0,0 +1,55 @@
...
+          commit-message: |
+            Update dashboards and alerts from Sunbeam https://opendev.org/openstack/sunbeam-charms/src/branch/main/charms/openstack-exporter-k8s/src
+
+            This commit was generated by https://github.com/canonical/snap-tempest-automation
```
aieri, line 47: needs updating

scripts/pull_dashboard_and_alert_rules.sh
aieri, line 1: optional: use a more generic script name in case we need to sync more than dashboard and alert rules in the future

scripts/pull_dashboard_and_alert_rules.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+# Clone the repository with sparse checkout
+git clone --sparse https://opendev.org/openstack/sunbeam-charms.git
```
aieri, line 4:
```suggestion
git clone --sparse --single-branch --depth=1 https://opendev.org/openstack/sunbeam-charms.git
```
just a small optimization

aieri (changes requested):
thank you this is looking really good! I have a few suggestions

aieri:
@Pjack the need for github apps is unfortunate but tbh I think we just have to get used to them for all sort of github automation work, so from my point of view they shouldn't be part of the cost/benefit analysis. And if we are only looking at the current implementation vs the existing custom script I think this proposal is simpler and more robust

aieri (changes requested):
ok just one more tiny nit: you now have

`scripts/sync_sunbeam.sh`
`workflows/sync-from-sunbeam.yaml`
`branch-name: chore/sunbeam-sync`

I'd choose one variant and keep names consistent

scripts/sync_sunbeam.sh
```diff
@@ -0,0 +1,16 @@
...
+# Set sparse checkout paths
+git sparse-checkout set charms/openstack-exporter-k8s/src/grafana_dashboards charms/openstack-exporter-k8s/src/prometheus_alert_rules
+
+# Copy the necessary files to the src directory
+cp -r charms/openstack-exporter-k8s/src/grafana_dashboards ../src/
+cp -r charms/openstack-exporter-k8s/src/prometheus_alert_rules ../src/
+
+# Clean up
+cd ..
+rm -rf sunbeam-charms
```
aieri, line 16: eeeeehhh `rm -rf`s give me shivers. If you **really** want to make the script cleanup after itself for the times when we may want to run it locally (because if it's run as part of the CI then the whole runner would be eventually torn down) I'd prefer using a temporary dir with a unique name and absolute paths.

See this link for a good example using traps: https://stackoverflow.com/a/34676160 (I'd just avoid the `-p` option and create the directory under `$TMPDIR`)

.github/workflows/pull-request.yaml
aieri: unresolving as this also applies to line 24

aieri: and line 2 of the workflow file

.github/workflows/sync-from-sunbeam.yaml
```diff
@@ -0,0 +1,46 @@
...
+          repositories: ${{ env.REPO_PATH }}
+      - name: Create pull request
+        uses: canonical/create-pull-request@main
+        with:
+          upsert: true  # Allow update PRs
+          github-token: ${{ steps.app_token.outputs.token }}
+          branch-name: chore/sunbeam-sync  # create new branch with the changes as the source branch for the PR
+          repository: ${{ env.REPO_OWNER }}/${{ env.REPO_PATH }}
+          title: "Update dashboards and alerts"
+          body: "This pull request was generated by https://github.com/canonical/openstack-exporter"
```
aieri, line 42: the repo name is incorrect

.github/workflows/sync-from-sunbeam.yaml
```diff
@@ -0,0 +1,46 @@
...
+          commit-message: |
+            Update dashboards and alerts from Sunbeam https://opendev.org/openstack/sunbeam-charms/src/branch/main/charms/openstack-exporter-k8s/src
+
+            This commit was generated by https://github.com/canonical/openstack-exporter
```
aieri, line 46: repo name is incorrect

aieri (changes requested):

scripts/sync-from-sunbeam.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/sh
+
+# the directory of the script
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+
+# the temp directory used, within $DIR
+# omit the -p parameter to create a temporal directory in the default location
+WORK_DIR=`mktemp -d -p "$DIR"`
```
aieri, line 8: why not create the dir under /tmp?

scripts/sync-from-sunbeam.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh
+
+# the directory of the script
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
```
samuelallan72, line 4: This script is looking pretty nice.  Just one thing: the shebang says /bin/sh which isn't guaranteed to be bash, but `BASH_SOURCE` is used here (bash-specific).  I'd recommend:

- change the shebang to /usr/bin/bash, /bin/bash, or /usr/bin/env bash (I'm not sure which is most recommended)
- make this script executable (git stores the executable bit of files, so you can be sure it will stay executable for new clones too)
- in the workflow yaml, call this script directly as an executable: `./scripts/sync-from-sunbeam.sh`

.github/workflows/sync-from-sunbeam.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+jobs:
+
+  sync-from-sunbeam:
+    name: Pull files from sunbeam
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout local repository
+        uses: actions/checkout@v4
+
```
samuelallan72, line 22: Remove trailing whitespace:

```suggestion

```

If you haven't already, I'd recommend configuring your editor or IDE to display trailing whitespace.  Eg. this is what it looks like in my neovim setup:

.github/workflows/sync-from-sunbeam.yaml
```diff
@@ -0,0 +1,46 @@
...
+          bash scripts/sync-from-sunbeam.sh
+
+      - uses: actions/create-github-app-token@v1
+        id: app_token
+        with:
+          app-id: ${{ secrets.APP_ID }}
+          owner: ${{ env.REPO_OWNER }}
+          private-key: ${{ secrets.APP_PEM }}
+          repositories: ${{ env.REPO_PATH }}
+      - name: Create pull request
```
samuelallan72, line 34: This should only create a PR if there are changes.  It looks like this will always create the PR?

.github/workflows/pull-request.yaml
```diff
@@ -21,6 +21,8 @@ jobs:
       tox-version: ""
       python-version: "3.10"
 
-  check-file-updates:
+  sync-from-sunbeam:
     name: Check updates for charm files
-    uses: ./.github/workflows/_check_file_update.yaml
+    uses: ./.github/workflows/sync-from-sunbeam.yaml
```
samuelallan72, line 26: I don't think we should call this workflow from the pull request checks workflow.  The sync workflow opens a pull request to make updates; it doesn't do any checks or report on status, so it doesn't add information to a pull request.

samuelallan72 (changes requested):

samuelallan72:
Comment on the design: currently we're going the route of having a workflow that does the syncing and auto-opens a PR with the changes.  I think this is nicely convenient, but I did see comments about maintaining a github app so the workflow can create signed commits.

If it's too much maintenance for signed commits + auto PRs, we could go another route:

- modify the workflow so it simply reports if there are changes
- add the workflow to the pull request workflow (so it alerts and blocks PRs if there are unsynced changes)
- also ensure the workflow is included in the weekly tests, so we're alerted periodically if there are changes
- if there are changes, one of us can manually run the sync script locally, and open a PR with the changes.

samuelallan72:
Also please update the contributing docs here: https://github.com/canonical/openstack-exporter-operator/blob/5f7aac7cc7c39589d76c8eb78516cedb7cb43b23/CONTRIBUTING.md#L57-L73 .  It should include information about which files/directories are managed by the syncing script, how to run the script locally, and what automation is in place.

scripts/sync-from-sunbeam.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# the temp directory used, within $DIR
```
aieri, line 6: no longer true now that the tmpdir is created under $TMPDIR or `/tmp`

scripts/sync-from-sunbeam.sh
aieri: there are other bashisms in this script too. Unless you really want this script to work under pure `sh`, setting the shebang to bash is the right way to go

aieri (changes requested):
one small suggestion

.github/workflows/sync-from-sunbeam.yaml
zxhdaze: Well according to https://github.com/canonical/create-pull-request by default, the action fails if no local changes are detected. I have enabled `ignore-no-changes` option to make sure the action to do nothing and exit successfully when there are no changes.

.github/workflows/sync-from-sunbeam.yaml
```diff
@@ -0,0 +1,49 @@
...
+        uses: canonical/create-pull-request@main
+        with:
+          upsert: true  # Allow update PRs
+          ignore-no-changes: false # Enabling this option allows the action to do nothing and exit successfully when there are no changes.
```
gabrielcocenza, line 38: Reading the [action config](https://github.com/canonical/create-pull-request?tab=readme-ov-file#inputsignore-no-changes) this should be set to true, isn't it?

zxhdaze: Right

PR #74, opened 2024-08-06
Set cache default option to True

gabrielcocenza:
Having the cache option set to True will make the exporter work out of the box. Right now if not changed, the time to scrape is bigger than 10 seconds and it fails.


gabrielcocenza:
Closing this PR because it's not possible to run github workflows from forked repos

PR #76, opened 2024-08-06
Set cache default option to True

gabrielcocenza:
Having the cache option set to True will make the exporter work out of the box. Right now if not changed, the time to scrape is bigger than 10 seconds and it fails.

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -49,7 +49,7 @@ config:
         Customizable based on the size of the OpenStack cluster.
         Format details: https://pkg.go.dev/time#ParseDuration
     cache:
-      default: false
+      default: true
       type: boolean
       description: |
         Enables the exporter cache globally. Refreshes at intervals of cache_ttl/2.
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -52,7 +52,7 @@ config:
       default: true
       type: boolean
       description: |
-        Enables the exporter cache globally. Refreshes at intervals of cache_ttl/2.
+        By default, enables the exporter cache globally. Refreshes at intervals of cache_ttl/2.
         If the cache is empty or expired, the response will be empty.
 
 
--- a/tests/integration/tests/charm_tests/openstack_exporter.py
+++ b/tests/integration/tests/charm_tests/openstack_exporter.py
@@ -83,6 +83,8 @@ class OpenstackExporterConfigTest(OpenstackExporterBaseTest):
         new_value = "random_ca"  # bad ssl_ca (will be reset shortly)
         old_value = model.get_application_config(APP_NAME).get(key)
         model.set_application_config(APP_NAME, {key: new_value})
+        # Change cache to false to not get previous scrape results
+        model.set_application_config(APP_NAME, {"cache": "false"})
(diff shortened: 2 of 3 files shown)


charmcraft.yaml
```diff
@@ -49,7 +49,7 @@ config:
         Customizable based on the size of the OpenStack cluster.
         Format details: https://pkg.go.dev/time#ParseDuration
     cache:
-      default: false
+      default: true
       type: boolean
       description: |
         Enables the exporter cache globally. Refreshes at intervals of cache_ttl/2.
```
samuelallan72, line 55: Could you update the description here to say that it defaults to true (enabled)?

samuelallan72:
Once merged, please be sure to update the charmhub docs ( https://charmhub.io/openstack-exporter ):

- the "enable the cache" page may need to be changed to "disable the cache" or "manage the cache feature" or something similar, since it's enabled by default now.
- the [metrics caching](https://charmhub.io/openstack-exporter/docs/cache) also will need updating.
- until it's on latest/stable, we probably should have a note about this changing in revision XX.

chanchiwai-ray (dismissed):

gabrielcocenza:
@samuelallan72 I've created a new issue #78 to track that.

Should we change the documentation after reaching the stable release?

samuelallan72 (approved):
Thanks, looks good. :)

> Should we change the documentation after reaching the stable release?

Agreed, that may be a good idea to avoid confusion. :+1:

chanchiwai-ray (approved):

jneo8 (commented):
question(non-blocking): I kind of lose why we want to enable cache as default, the upstream exporter default is false.

gabrielcocenza:
@jneo8 
> question(non-blocking): I kind of lose why we want to enable cache as default, the upstream exporter default is false.

The motivation behind it it's because the scrape was failing, at least on serverstack (I didn't try on partner cloud or other environment that is more realistic with production) and it's not a good user experience to try to figure out why it's not working. Having it set to True we kind guarantee that will start working out of the box

jneo8:
> The motivation behind it it's because the scrape was failing, at least on serverstack (I didn't try on partner cloud or other environment that is more realistic with production) and it's not a good user experience to try to figure out why it's not working. Having it set to True we kind guarantee that will start working out of the box

My concern is about the meaning of the metrics become a little bit different after you open the cache. This is also a major concern from others so we decide to not enable it as default and provide the same experience as upstream.

I am ok to enable cache as default, if there are more concern from others then we can discuss if we want to set it back.

tests/integration/tests/charm_tests/openstack_exporter.py
```diff
@@ -107,6 +109,10 @@ def test_configure_ssl_ca(self):
         results = model.run_on_leader(APP_NAME, command)
         self.assertEqual(int(results.get("Code", "-1")), 0)
 
+        # Enable cache again
+        model.set_application_config(APP_NAME, {"cache": "true"})
+        model.block_until_all_units_idle()
```
rgildein, line 114: I believe this should be done by clean up function, so it will be set to true even if tests failed.

PR #80, opened 2024-08-08
Change grafana-agent to use stable channel on functional tests

gabrielcocenza:
Now that grafana-agent has a stable release, it's better to change our workflow to use it to avoid the CI breaking on the edge channel.

E.g: canonical/grafana-agent-operator#165


zxhdaze (approved):

samuelallan72 (approved):
Thanks, good point

jneo8 (approved):

PR #83, opened 2024-08-09
Pin charmcraft to v2 for releases

samuelallan72:
The charmcraft.yaml isn't yet updated to work properly with charmcraft 3


chanchiwai-ray (approved):

Pjack (approved):

jneo8 (approved):

samuelallan72:
Merging without func tests because this only touches the release.yaml (and that only runs from main).  Thanks for the reviews. :)

PR #84, opened 2024-08-09
Inline bootstack actions func workflow

samuelallan72:
This lets us simplify the workflow,
gives us more control over testing in the future,
and also lets us pin charmcraft to v2
(this charm doesn't official support charmcraft 3).


.github/workflows/check.yaml
```diff
@@ -0,0 +1,118 @@
...
+        run: |
+          mkdir -p $HOME/.local/share/juju
+          # TODO: can we remove the base64 encoding?
+          echo ${{ secrets.JUJU_CONTROLLERS_YAML }} | base64 -d > $HOME/.local/share/juju/controllers.yaml
+          echo ${{ secrets.JUJU_ACCOUNTS_YAML }} | base64 -d > $HOME/.local/share/juju/accounts.yaml
+
+      - name: Configure latest lxd
+        uses: canonical/setup-lxd@2aa6f7caf7d1484298a64192f7f63a6684e648a4  # latest main at time of writing
+        with:
+          channel: latest/stable
```
chanchiwai-ray, line 82: What is the step?

.github/workflows/check.yaml
```diff
@@ -0,0 +1,118 @@
...
+           juju show-controller soleng-ci-ctrl-34 &>/dev/null
+
+      - name: Show juju information
+        run: |
+          juju version
+          juju controllers | grep Version -A 1 | awk '{print $9}'
+
+      - name: Show snap information
+        run: |
+          snap list
```
chanchiwai-ray, line 113: Also this

samuelallan72: This is just debugging information; I ported it across from the bootstack actions.  I can remove it, it's probably not that useful.

.github/workflows/check.yaml
samuelallan72: the bootstack actions was setting up lxd.  However looking now, I agree - I don't know what the step is for.  I think it's not needed since we're using an external controller.

samuelallan72: Removed

.github/workflows/check.yaml
samuelallan72: Removed

chanchiwai-ray:
thanks, that looks like bootstack action has some redundant steps

gabrielcocenza:
Looking good, but shouldn't this be applied via terraform?

samuelallan72:
> thanks, that looks like bootstack action has some redundant steps

@chanchiwai-ray could you clarify which steps you're referring to?

> Looking good, but shouldn't this be applied via terraform?

@gabrielcocenza the workflow I've found most efficient so far working on these workflows is to add them directly to the repository first so I can iterate quickly with feedback from running the workflows.  Then once it's approved and merged, I can port it to the automation repo.

chanchiwai-ray:
> @chanchiwai-ray could you clarify which steps you're referring to? 

The lines you need to removed, but it's done

.github/workflows/check.yaml
```diff
@@ -0,0 +1,109 @@
...
+           # Workaround to avoid missing cookies bug
+           # https://github.com/search?q=repo%3Ajuju%2Fpython-libjuju%20cookies_for_controller&type=code
+           juju show-controller soleng-ci-ctrl-34 &>/dev/null
+
+      - name: Show juju information
+        run: |
+          juju version
+          juju controllers | grep Version -A 1 | awk '{print $9}'
+
+      - name: Run tests
```
chanchiwai-ray, line 106: Ah so this step depends on setup-lxd 😞

samuelallan72: Ah of course, charmcraft needs lxd to build the charm.

chanchiwai-ray (approved):

jneo8 (approved):

.github/workflows/check.yaml
```diff
@@ -0,0 +1,114 @@
...
+
+      - name: install external controller required packages
+        run: |
+           sudo snap install charmcraft --channel 2.x/stable --classic
+           sudo snap install juju --channel 3.4/stable --classic
+           sudo --preserve-env=http_proxy,https_proxy,no_proxy pip3 install tox
+           juju switch soleng-ci-ctrl-34
+           # Workaround to avoid missing cookies bug
+           # https://github.com/search?q=repo%3Ajuju%2Fpython-libjuju%20cookies_for_controller&type=code
+           juju show-controller soleng-ci-ctrl-34 &>/dev/null
```
rgildein, line 104: I think it would be better if we make these parameters (3.4/stable and soleng-ci-ctrl-34) as matrix? WDYT?

samuelallan72: I don't see any need.  If we want to test with multiple channels in the future, then we can add it.

Also note that we're moving away from the self-hosted juju controller soon, so we won't have any need to test on multiple controllers.

rgildein: Only reason is that if we need to change controller or juju version, it's more clear for anyone to do so at single place.

PR #85, opened 2024-08-09
Use the charmed-openstack-exporter snap

gabrielcocenza:
Upstream can take some time until releases a new snap and this is blocking the workflow.

This change from using the golang-openstack-exporter (upstream) to use the charmed-openstack-exporter that is packaging the upstream code and release in the snapcraft store.

With this change the charm automatically uses the snap from the snapstore and removes the upstream snap and/or snap as a resource.

The resource can still be used as a way to test customized snap or an alternative solution for airgapped environments


aieri (dismissed):

Makefile
```diff
@@ -28,15 +28,12 @@ clean:
 	@rm -f ${PROJECTPATH}/${CHARM_NAME}*.charm
 	@charmcraft clean
 	@echo "Remove download snap"
-	@rm -f ${PROJECTPATH}/golang-openstack-exporter*.snap
+	@rm -f ${PROJECTPATH}/charmed-openstack-exporter*.snap
```
samuelallan72, line 31:
```suggestion
```

We should remove all references to the local snap and the `download-snap` target; these were temporary things to work with our custom built snap.

I think we also shouldn't delete a file that the user would have manually put there; it would be unexpected.

Makefile
```diff
@@ -28,15 +28,12 @@ clean:
...
 
 build: clean
 	@echo "Building charm"
 	@charmcraft -v pack
 
-download-snap:
-	wget -q https://github.com/canonical/openstack-exporter-operator/releases/download/rev2/golang-openstack-exporter_amd64.snap -O ./golang-openstack-exporter_amd64.snap
-
 integration: build download-snap
```
samuelallan72, line 37:
```suggestion
integration: build
```

src/service.py
```diff
@@ -49,15 +51,21 @@ def configure(self, snap_config: dict[str, Any]) -> None:
...
+    If resource is passed it will have preference over the one in the snapcraft store.
+
     Raises an exception on error.
     """
     try:
         logger.debug("installing snap.")
-        logger.debug("fetching from resource.")
-        snap_client = snap.install_local(resource, dangerous=True)
+        if resource:
+            logger.debug("fetching %s from resource.", SNAP_NAME)
```
samuelallan72, line 64:
```suggestion
            logger.debug("installing %s snap from resource.", SNAP_NAME)
```

"fetching" doesn't really match what is happening.

src/service.py
```diff
@@ -49,15 +51,21 @@ def configure(self, snap_config: dict[str, Any]) -> None:
...
+            snap_client = snap.install_local(resource, dangerous=True)
+        else:
+            logger.debug("fetching %s from snapcraft store", SNAP_NAME)
```
samuelallan72, line 67:
```suggestion
            logger.debug("installing %s from snapcraft store", SNAP_NAME)
```

tests/integration/tests/bundles/overlays/local-charm-overlay.yaml.j2
```diff
@@ -1,5 +1,7 @@
 applications:
   {{ charm_name }}:
     charm: {{ CHARM_LOCATION }}
+    {% if CHARM_SNAP_LOCATION %}
     resources:
       openstack-exporter: {{ CHARM_SNAP_LOCATION }}
+    {% endif %}
```
samuelallan72, line 7: Please document this in CONTRIBUTING.md (that it will use a local resource if there is a particularly named snap file in the root of the repo, that it relies on env vars set from the makefile).

src/service.py
```diff
@@ -49,15 +51,21 @@ def configure(self, snap_config: dict[str, Any]) -> None:
         self.snap_client.set(snap_config, typed=True)
 
 
-def snap_install(resource: str) -> SnapService:
+def snap_install(resource: Optional[str]) -> SnapService:
```
samuelallan72, line 54: If a user attaches a snap resource after deployment, will the snap resource be installed, replacing the snap from the store?  Conversely, if a user removes the snap resource, will the snap revert back to the one from the store?  I think this would be desired behaviour.

samuelallan72 (changes requested):

Makefile
chanchiwai-ray: +1, `download-snap` should be removed, and when it's removed line 30-31 should also be removed

gabrielcocenza: I forgot to remove it before. Thanks

src/service.py
gabrielcocenza: I took quite some time trying to figure out how I could detect that a snap was installed from a local file and not from snapstore. Long story short there is no clear information about this. The only thing that I found is that when you install locally it will show an  `x` on the revision. This [post](https://forum.snapcraft.io/t/how-to-check-if-a-snap-is-installed-in-dangerous-mode/3968) confirms that.

It's not possible to remove a snap resource from the charm, what you can do is to pass a resource. When that is done, it was failing because the snap couldn't refresh with a local installation. In order to solve this I'm detecting if there is `x` in the version and in that case I remove the snap to be able to install from the snapstore again.

samuelallan72: thanks for digging into this! :)

CONTRIBUTING.md
```diff
@@ -31,11 +31,14 @@ make update-charm-libs          # update charm's libraries
...
 make sync-dashboards            # update the dashboards from upstream
 make clean                      # remove unneeded files
-make download-snap              # download snap release from github release assets
 make build                      # build the charm
 make integration                # run the tests defined in the integration subdirectory
 ```
 
+Note: You can use a local snap exporter to use in the integration test by placing it in the root of
+the repo and naming the snap file with the `charmed-openstack-exporter` prefix. When this happens,
+the `CHARM_SNAP_LOCATION` environment variable will detect the file and attach it as a resource.
```
samuelallan72, line 40: The wording could be improved:

```suggestion
the Makefile will detect the file and save the path to the `CHARM_SNAP_LOCATION` environment variable.
With this environment variable set, the zaza test will attach this file as a resource.
```

samuelallan72 (dismissed):
Looks good, thank for working on this. :)  I just flagged a docs clarification.

chanchiwai-ray (approved):

gabrielcocenza:
@samuelallan72 @chanchiwai-ray after talking with Americas, we decided to remove the possibility of using a custom snap as resource, so the code is reflecting this now. Moreover, I remembered that it's necessary to remove the upstream snap to not conflict with the new one. The `snap.remove` can be used even if the snap is not present so that is why it's outside the try except block

samuelallan72 (changes requested):
I'm not sure removing support for the snap resource is a good idea, so leaving my review in 'request changes' until I understand the rationale.

src/service.py
```diff
@@ -8,6 +8,9 @@
 
 logger = getLogger(__name__)
 
+SNAP_NAME = "charmed-openstack-exporter"
+UPSTREAM_SNAP = "golang-openstack-exporter"
```
jneo8, line 12: **question:** 

What is the different? To me it's two different snap name and it make me confuse.

src/service.py
```diff
@@ -49,15 +52,25 @@ def configure(self, snap_config: dict[str, Any]) -> None:
...
+def snap_install(resource: Optional[str]) -> SnapService:
+    """Install the snap, and return the snap service.
+
+    Before installing the snap, it will remove the canonical and the upstream snap to ensure that
+    the snap installed is not a resource. Removing the golang exporter also ensure that there is
+    no conflict using the one from Canonical.
 
     Raises an exception on error.
     """
+    snap.remove([SNAP_NAME, UPSTREAM_SNAP])
```
jneo8, line 64: **thought (non-blocking):** 

Instead of remove upstream snap for the user, maybe a blocked status is more suitable?

(Just want to make sure it doesn't have any side effort for the user, feel free to resolve it.)

src/service.py
```diff
@@ -88,7 +101,7 @@ def workaround_bug_268() -> None:
     https://github.com/openstack-exporter/openstack-exporter/issues/268
     """
     logger.info("Adding service override to workaround bug 268")
-    dir = "/etc/systemd/system/snap.golang-openstack-exporter.service.service.d"
+    dir = "/etc/systemd/system/snap.charmed-openstack-exporter.service.service.d"
     os.makedirs(dir, exist_ok=True)
     with open(f"{dir}/bug_268.conf", "w") as f:
         f.write("[Service]\nEnvironment=OS_COMPUTE_API_VERSION=2.87\n")
```
jneo8, line 107: **nitpick (blocking):** 

Suggest to move these line to a independent function.

jneo8 (changes requested):
Thanks for making this happen, finally we have the snap. Some feedback.

chanchiwai-ray:
> after talking with Americas, we decided to remove the possibility of using a custom snap as resource, so the code is reflecting this now. Moreover, I remembered that it's necessary to remove the upstream snap to not conflict with the new one. The snap.remove can be used even if the snap is not present so that is why it's outside the try except block

Just want to know the story, why suddenly decide not to use resource. The resource metadata can't be taken down from charmhub, and may not be a good user experience

src/service.py
samuelallan72: Agreed, perhaps naming it `OLD_SNAP_NAME` or something to make it clearer how it's used?

src/service.py
samuelallan72: I don't mind this method - it's nice and seamless for the end user.

samuelallan72: But this logic means we'll be removing and re-installing the snap each time, which could lead to unecessary service restarts?

src/service.py
```diff
@@ -49,15 +52,25 @@ def configure(self, snap_config: dict[str, Any]) -> None:
...
 
     Raises an exception on error.
     """
+    snap.remove([SNAP_NAME, UPSTREAM_SNAP])
     try:
-        logger.debug("installing snap.")
-        logger.debug("fetching from resource.")
-        snap_client = snap.install_local(resource, dangerous=True)
+        logger.debug("installing %s from snapcraft store", SNAP_NAME)
+        if resource:
```
samuelallan72, line 67: If we're dropping support for resource, can we remove this logic altogether?  I heard there was a limitation related to charmhub about removing resource metadata from a charm, but it's not confirmed.  What happens if we remove the resource from charm metadata?

If we cannot remove the resource from the charm metadata now that the charm is published, then I'm in favour of retaining support.  Otherwise it's confusing if the charm has a resource that isn't used.

src/service.py
samuelallan72: They are in an independent function already - it's in `workaround_bug_268()`.

samuelallan72 (changes requested):

samuelallan72:
> I'm not sure removing support for the snap resource is a good idea, so leaving my review in 'request changes' until I understand the rationale.

Following up on this: we discussed internally, and the rationale for removing it was to simplify the charm (user experience and code logic).  We discussed airgapped environments, but it was put forward that [snap store proxy](https://docs.ubuntu.com/snap-store-proxy/en/install) would be used as the recommended method for handling this.

src/service.py
gabrielcocenza: They are different snaps, but have the same code base. We need to ensure to remove before installing the one that we are handling. Is changing the name enough for you @jneo8 ?

src/service.py
gabrielcocenza: I don't see any side effect of removing the upstream snap and installing the one that we are handling.

This will be called on charm upgrade or when installing so it's kind expected to restart the service. I don't see a big issue particularly .

src/service.py
gabrielcocenza: My gut feeling is that it's not possible like when you add a charm config, but I didn't find documentation about this. I've asked in the charm development room in matrix and let's see.

If it's possible I agree with you

src/service.py
samuelallan72: > I don't see any side effect of removing the upstream snap and installing the one that we are handling.

Agreed, but note that we're also removing the one we're handling here too.  Not sure if that's desired.

src/service.py
samuelallan72: > like when you add a charm config

Oh it's not possible to remove charm config too?  Do you know of docs that cover this?  That doesn't make sense.

src/service.py
chanchiwai-ray: No, they are not the same from the snap store perspective; the one in snap store for [golang-openstack-exporter](https://snapcraft.io/golang-openstack-exporter) was like 3 years ago. And it does not have a service for the charm to manage. It's just a cli tool

src/service.py
```diff
@@ -88,7 +101,7 @@ def workaround_bug_268() -> None:
     https://github.com/openstack-exporter/openstack-exporter/issues/268
     """
     logger.info("Adding service override to workaround bug 268")
-    dir = "/etc/systemd/system/snap.golang-openstack-exporter.service.service.d"
+    dir = "/etc/systemd/system/snap.charmed-openstack-exporter.service.service.d"
```
chanchiwai-ray, line 104: nit: it's better to use `SNAP_NAME` here

src/service.py
chanchiwai-ray: or you mean the resource of `golang-openstack-exporter`, if so, please ignore my comment

src/service.py
chanchiwai-ray: One side effect of leaving it around is that they might use the same port for the exporter. Can we make the logic of detecting old exporter and removing old exporter in a single function, and add comment saying that this function can be removed in the future when we fully adopt the "new" exporter. I see this issue as a transient issue: nobody will ever notice that in the future

chanchiwai-ray: Something like and we wouldn't need `UPSTREAM_NAME` constant
```python
def warn_and_remove_local_snap_resource():
  try:
    snap.remove(["golang-openstack-exporter"])
  except snap.SnapNotFoundError:
    pass
  if resource:
     logger.warning(
          "Local resources cannot be used. Installing the exporter %s from snap store",
          SNAP_NAME,
      ) 
```

gabrielcocenza: I did something similar to this, but we don't need to worry about this SnapNotFoundError because the [lib](https://github.com/canonical/openstack-exporter-operator/blob/main/lib/charms/operator_libs_linux/v2/snap.py#L573-L578) already handle when it's not present.

src/service.py
```diff
@@ -67,6 +72,14 @@ def snap_install(resource: str) -> SnapService:
         return SnapService(snap_client)
 
 
+def remove_upstream_snap() -> None:
+    """Remove the old snap from upstream to not override with the charmed-openstack-exporter."""
```
samuelallan72, line 76:
```suggestion
    """Remove the old snap from upstream to not conflict with the charmed-openstack-exporter."""
```

src/service.py
```diff
@@ -67,6 +72,14 @@ def snap_install(resource: str) -> SnapService:
...
+    try:
+        snap.remove(["golang-openstack-exporter"])
+    except snap.SnapError as e:
+        logger.error("failed to remove snap: %s", str(e))
```
samuelallan72, line 80: I don't think we should ignore this error; if the old snap fails to remove, the charm should block.  It's not a good state to have both installed at once.

src/charm.py
```diff
@@ -4,41 +4,34 @@
 # See LICENSE file for licensing details.
 """OpenStack Exporter Operator.
 
-This charm provide golang-openstack-exporter snap as part of the Charmed
+This charm provide charmed-openstack-exporter snap as part of the Charmed
```
samuelallan72, line 7:
```suggestion
This charm provides charmed-openstack-exporter as part of the Charmed
```

samuelallan72 (changes requested):

src/service.py
samuelallan72: Ok, it sounds like it's fine to remove resources or config.  The only issue is that it can make some things non-backward compatible (eg. if users were using a custom snap as a resource, now they can't)

samuelallan72:
it looks pretty close now, thanks!  I just noted a couple of nits and a query about errors removing the old/upstream snap.

jneo8:
> Following up on this: we discussed internally, and the rationale for removing it was to simplify the charm (user experience and code logic). We discussed airgapped environments, but it was put forward that [snap store proxy](https://docs.ubuntu.com/snap-store-proxy/en/install) would be used as the recommended method for handling this.

> It does remove the convenience of testing with a custom built snap (eg. if we've working on a new feature or fix in openstack-exporter), but we can workaround that (manually ssh in to the unit and install it).

> Something I'm not sure about is the future of the snap resource in charm metadata going forward. I've heard tell that we can't remove this now the charm is published on charmhub. Can we confirm that? If we can cleanly remove the resource from metadata, then I'm fine with going forward with removing it. However, if it's stuck in metadata forever, I'm in favour of retaining support, to avoid confusion in the user experience.

(non-blocking) I am +1 to keep the snap resource support.

chanchiwai-ray (approved):

nobuto-m:
> With this change the charm automatically uses the snap from the snapstore and removes the upstream snap and/or snap as a resource.
> 
> The resource is removed completely and won't be possible to use it as a resource

The charm declared GA already, and this simply breaks the backward compatibility after that. Can't this part of the change be a separate review?

samuelallan72:
> this simply breaks the backward compatibility after that.

@nobuto-m could you help us figure out exactly what breaks in the situation of removing the resource?  We asked the charm teams and it sounds like the resource will simple silently drop from the charm and the charm will switch from using the custom snap from resource to installing it from the snap store.

nobuto-m:
> > this simply breaks the backward compatibility after that.
> 
> @nobuto-m could you help us figure out exactly what breaks in the situation of removing the resource? We asked the charm teams and it sounds like the resource will simple silently drop from the charm and the charm will switch from using the custom snap from resource to installing it from the snap store.

One of the motivations to use a snap resource is for an air-gapped environment. I understand that the ideal solution is to use air-gapped mode of snap-store-proxy. However, when upgrading the charm for such an existing environment, does `juju refresh` just work? I don't think so. That's considered as a major breaking change in that context so separating out the change to drop the resource support would be the best thing for the time being.

src/service.py
```diff
@@ -49,15 +51,18 @@ def configure(self, snap_config: dict[str, Any]) -> None:
...
+    older versions of this charm.
 
     Raises an exception on error.
     """
+    remove_upstream_snap()
     try:
-        logger.debug("installing snap.")
-        logger.debug("fetching from resource.")
-        snap_client = snap.install_local(resource, dangerous=True)
+        snap_client = snap.add(SNAP_NAME, channel="latest/stable")
```
nobuto-m, line 64: hardcoding a channel name is generally not a good idea since it will prevent testing with latest/edge or latest/candidate for example.

src/service.py
```diff
@@ -49,24 +51,71 @@ def configure(self, snap_config: dict[str, Any]) -> None:
...
     Raises an exception on error.
     """
+    remove_upstream_snap()
     try:
-        logger.debug("installing snap.")
-        logger.debug("fetching from resource.")
-        snap_client = snap.install_local(resource, dangerous=True)
+        if resource:
+            logger.debug("installing %s from resource.", SNAP_NAME)
+            # installing from a resource if installed from snap store previously is not problematic
```
samuelallan72, line 69: Thanks for testing this and for the useful comments in this function :+1:

samuelallan72 (dismissed):
Thanks @gabrielcocenza , I think the requirements are clear now and this is looking good.  I'm not sure about using stored state though, and I don't think it will be necessary.

src/charm.py
```diff
@@ -231,6 +236,13 @@ def _on_collect_unit_status(self, event: ops.CollectStatusEvent) -> None:
 
         event.add_status(ActiveStatus())
 
+    def snap_channel_changed(self) -> bool:
```
jneo8, line 239: **nitpick (blocking):** 

The snap version information shouldn't be stored in the local stored. It's the information you can get in live time.

src/service.py
```diff
@@ -49,24 +51,71 @@ def configure(self, snap_config: dict[str, Any]) -> None:
         self.snap_client.set(snap_config, typed=True)
 
 
-def snap_install(resource: str) -> SnapService:
-    """Install or refresh the snap, and return the snap service.
+def snap_install(resource: Optional[str], channel: str) -> SnapService:
```
jneo8, line 54: **nitpick (non-blocking):** 

I have question about this design. The factory function to get the service and install should be separate.

One with side-effect to install the snap. Another is getting the snap service encapsulation, which don't have side-effect.

(Feel free to refactor this part in another PR)

src/service.py
```diff
@@ -49,24 +51,71 @@ def configure(self, snap_config: dict[str, Any]) -> None:
...
+
+    Before installing the snap, it will try to remove the upstream snap that could be installed on
+    older versions of this charm.
+
+    If the channel is changed, the snap.add method is able to identify and refresh the charm to the
+    new channel.
 
     Raises an exception on error.
     """
+    remove_upstream_snap()
```
jneo8, line 65: **issue (blocking):** 

If the resource is empty, the upstream snap will be also removed.

src/service.py
```diff
@@ -49,24 +51,71 @@ def configure(self, snap_config: dict[str, Any]) -> None:
...
+    installed by a local file. The way to return to the one from snapstore is by passing an empty
+    file. In such scenario, the local installation will be removed to be able to install from the
+    snapstore.
+    """
+    snap_cache = snap.SnapCache()
+    o7k_exporter = snap_cache[SNAP_NAME]
+    if o7k_exporter.present and "x" in o7k_exporter.revision:
+        logger.info("removing local resource snap before installing from snapstore")
+        try:
+            print("Trying to remove")
```
jneo8, line 112: **nitpick (blocking):** 

Change to log?

jneo8 (changes requested):
Thanks for the change. Some nitpick comments.

chanchiwai-ray (commented):
The logic with `resource` and `snap_channel` is a bit convoluted now and should be carefully tested in functional test (can be another PR is that's too big of a change) to avoid subtle bugs.

- If non-empty resource is present, what will happen if the user change `snap_channel` config. Currently, it will 
  1 `snap_install` 
  2 `remove_upstream_snap`
  3 `snap.install_local` ...
- If a empty resource is present, what will happen if the user change `snap_channel` config. Currently, it will
  1. `snap_install`
  2. `remove_upstream_snap`
  3. `remove_snap_as_resource`
  4. `snap.add`
- If user want to use snap from snap store to replace the resource, does the step will be performed correctly in the charm code (upload an empty resource + set desired `snap_channel`). Currently, it will
  1. `snap_install` (install hook)
  2. `remove_upstream_snap`
  3. `remove_snap_as_resource`
  4. `snap.add`
  5. `snap_install` (config_change hook, if snap_channel changed)
  6.  `remove_upstream_snap`
  7. `remove_snap_as_resource`
  8. `snap.add`

src/service.py
gabrielcocenza: That is intentional. Not sure if it will have people using the snap form the snapstore, but I don't see an issue of removing it independently if it was used as resource or from the store

src/service.py
gabrielcocenza: We already have the `get_installed_snap_service`, so I think we can remove to return the `SnapService` and this function is only responsible for installing or refresh. Moreover we are never using the response

charmcraft.yaml
```diff
@@ -71,16 +78,8 @@ resources:
...
-      since the supported snap isn't published on the snap store.
-      When it is published, the charm will use the snap from the store,
-      and a user can provide a local snap to use instead.
-
-      This override may be useful for testing or in environments with network restrictions.
+      This is the software that implements actual metrics collection and endpoint for prometheus
+      scraping. The snap resource on Charmhub is intended to be an empty file. The charm will
+      ignore an empty file resource and install from the snap store, so that the resource can be
+      used as an override by the user, if they wish to provide a local snap or if it's an airgapped
+      environment where snap-proxy is not a possible solution.
```
aieri, line 85: rewording suggestion
```suggestion
      This is the software that implements actual metrics collection and an endpoint for prometheus
      scraping. The snap resource on Charmhub is intended to be an empty file. The charm will
      ignore an empty file resource and normally install the snap from the snap store, but a custom snap can also be provided here if needed.
```
(I would not mention snap-proxy not being a possible solution since it's supposed to be a possible solution :))

src/service.py
aieri, line 54: non-blocking: I know these functions were already here, but they do look odd. Given the mostly OO design of this code these utility functions feel more they would belong as methods of a Snap class.

aieri: reworking these into a Snap class (which could be done in a separate PR) could also help resolve the awkwardness of having both a resource path and a channel in the same signature (channel and path would be class attributes and install would inspect them to determine how to behave)

src/service.py
aieri: I think this is correct, if the resource is empty we should switch from golang-openstack-exporter to charmed-openstack-exporter

src/service.py
```diff
@@ -49,24 +51,71 @@ def configure(self, snap_config: dict[str, Any]) -> None:
...
         return SnapService(snap_client)
 
 
+def remove_upstream_snap() -> None:
+    """Remove the old snap from upstream to not conflict with the charmed-openstack-exporter.
+
+    Raises an exception on error.
+    """
+    try:
+        snap.remove(["golang-openstack-exporter"])
```
aieri, line 92: does this function raise an error if golang-openstack-exporter is not present?

aieri (changes requested):
a few small comments

src/service.py
gabrielcocenza: No, the lib is smart in that sense. It's just an extra cautious if an unknown exception raises

aieri (approved):

chanchiwai-ray (approved):
I think this is good enough for this iteration, thanks!

src/charm.py
```diff
@@ -15,30 +15,26 @@
...
-SNAP_NAME = "golang-openstack-exporter"
-
 # Snap config options global constants
 # This is to match between openstack-exporter and the entry in clouds.yaml
 CLOUD_NAME = "openstack"
 # store the clouds.yaml where it's easily accessible by the openstack-exporter snap
 # This is the SNAP_COMMON directory for the exporter snap, which is accessible,
 # unversioned, and retained across updates of the snap.
-OS_CLIENT_CONFIG = "/var/snap/golang-openstack-exporter/common/clouds.yaml"
+OS_CLIENT_CONFIG = f"/var/snap/{SNAP_NAME}/common/clouds.yaml"
```
nobuto-m, line 37: What happens in an upgrade scenario? i.e. have you tested possible upgrade scenarios?

Let's say:

- the charm was deployed with a local resource of golang-openstack-exporter snap initially
- the charm gets upgraded to the charm revision with this code
- the resource was attached initially with the golang-openstack-exporter snap and it's still there
- new charm revision skips the installation of the new charmed-openstack-exporter snap from the store (I might be wrong here)

Can the old golang-openstack-exporter snap read the new path with the snap confinement still?

gabrielcocenza: Good observation @nobuto-m .

The current implementation is not prepared to have the `golang-openstack-exporter` attached after the upgrade because it will remove but install it again. I will open a new PR to handle this situation.

Thanks

gabrielcocenza: Please take a look at the #93 when you have some time to solve this observation of yours.

PR #91, opened 2024-08-20
Add extra-bindings for all OpenStack endpoint interface types

### LINKED ISSUE #90: extra-bindings has gone
Assuming this charm is a rewrite of https://charmhub.io/prometheus-openstack-exporter, the previously supported `extra-bindings` (https://git.launchpad.net/charm-prometheus-openstack-exporter/tree/src/metadata.yaml#n30) has gone without any commit history.

There were reasons to have such metadata especially this charm handles OpenStack with admin/internal/public endpoints.

This is a blocker to use the new charm since it doesn't cover the existing deployment scenarios, i.e. complex enterprise networking where a routing from admin network to public network is not allowed.

aieri:
Add extra-bindings to the charm metadata to allow users to request legs
in all OpenStack networks spaces. This may be necessary in case the
OpenStack API endpoints to be queried are not available on the same
network space as the one used by the credentials endpoint.

While the charm doesn't explicitly use the bindingsi - meaning that the
same result could be achieved with machine space contraints - having
these specified as application bindings is more explicit.

This is effectively the same solution used in the previous exporter
charm, see
https://git.launchpad.net/charm-prometheus-openstack-exporter/commit/?id=99692c082eadbf5195bb683e92a73428603efbc6

Closes #90

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -28,9 +28,9 @@ provides:
 # network spaces. Needed when APIs are not exposed in the same network as the one bound
 # to the credentials endpoint
 extra-bindings:
-  admin-interface:
-  public-interface:
-  internal-interface:
+  admin:
+  public:
+  internal:
 
 config:
   options:
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -24,6 +24,14 @@ provides:
   cos-agent:
     interface: cos_agent
 
+# These bindings can be used to explicitly request interfaces in all the OpenStack
+# network spaces. Needed when APIs are not exposed in the same network as the one bound
+# to the credentials endpoint
+extra-bindings:
+  admin-interface:
+  public-interface:
+  internal-interface:
+
 config:
   options:
     port:
@@ -84,8 +92,3 @@ resources:
       and a user can provide a local snap to use instead.
 
       This override may be useful for testing or in environments with network restrictions.
-
-extra-bindings:
-  admin-interface:
-  public-interface:
-  internal-interface:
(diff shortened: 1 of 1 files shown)


charmcraft.yaml
```diff
@@ -24,6 +24,14 @@ provides:
   cos-agent:
     interface: cos_agent
 
+# These bindings can be used to explicitly request interfaces in all the OpenStack
+# network spaces. Needed when APIs are not exposed in the same network as the one bound
+# to the credentials endpoint
+extra-bindings:
+  admin-interface:
+  public-interface:
+  internal-interface:
```
nobuto-m, line 33: Is there any strong reason not to use the same naming convention with the previous charm and other OpenStack charms?
- https://git.launchpad.net/charm-prometheus-openstack-exporter/commit/?id=99692c082eadbf5195bb683e92a73428603efbc6
- https://opendev.org/openstack/charm-keystone/src/branch/master/metadata.yaml#L15-L18

nobuto-m (changes requested):

charmcraft.yaml
aieri: not a strong one; I am proposing the `-interface` suffix to be a bit more clear that there's nothing special about these bindings, it's just a trick to ensure the unit is created with extra interfaces

nobuto-m: To me, it's an unnecessary diff. This charm resides in the Charmed OpenStack ecosystem so I don't see a necessity to diverge from the standard naming convention.

samuelallan72: Agreed, let's stick to convention to avoid confusion.  We can add info to the charmhub docs about the purpose of the extra bindings if needed.

aieri: sure

samuelallan72 (approved):
Thanks, LGTM

chanchiwai-ray (approved):

nobuto-m (approved):

PR #93, opened 2024-08-23
Check to see if it's not using the upstream snap

gabrielcocenza:
- block units if upstream snap is installed via resource
- not run config-changed when upstream resource is used
- added func test showing that units go into blocked state if upstream resource is added.
- used snap cli to query snaps installed on functional tests

Closes: #94

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -10,7 +10,7 @@ OpenStack deployment.
 
 import logging
 import os
-from typing import Any, Optional
+from typing import Any
 
 import ops
 import yaml
@@ -19,11 +19,10 @@ from charms.operator_libs_linux.v2.snap import SnapError
 from ops.model import (
     ActiveStatus,
     BlockedStatus,
-    ModelError,
     WaitingStatus,
 )
 
-from service import SNAP_NAME, get_installed_snap_service, snap_install_or_refresh
+from service import SNAP_NAME, SnapResource, get_installed_snap_service, snap_install_or_refresh
 
 logger = logging.getLogger(__name__)
 
@@ -131,29 +130,11 @@ class OpenstackExporterOperatorCharm(ops.CharmBase):
                     return data
         return {}
 
-    def get_resource(self) -> Optional[str]:
-        """Return the path-to-resource or None if the resource is empty.
-
-        Fetch the charm's resource and check if the resource is an empty file
-        or not. If it's empty, return None. Otherwise, return the path to the
-        resource.
-        """
-        try:
-            snap_path = self.model.resources.fetch(RESOURCE_NAME).absolute()
-        except ModelError:
-            logger.debug("cannot fetch charm resource")
-            return None
-
-        if not os.path.getsize(snap_path) > 0:
-            logger.debug("resource is an empty file")
-            return None
-
-        return snap_path
-
     def install(self) -> None:
         """Install the necessary resources for the charm."""
         try:
-            snap_install_or_refresh(self.get_resource(), self.model.config["snap_channel"])
+            resource = SnapResource(RESOURCE_NAME, self.model)
+            snap_install_or_refresh(resource, self.model.config["snap_channel"])
         except SnapError:
             self.model.unit.status = BlockedStatus(
                 "Failed to remove/install openstack-exporter snap"
--- a/src/charm.py
+++ b/src/charm.py
@@ -15,14 +15,18 @@ from typing import Any
 import ops
 import yaml
 from charms.grafana_agent.v0.cos_agent import COSAgentProvider
-from charms.operator_libs_linux.v2.snap import SnapError
 from ops.model import (
     ActiveStatus,
     BlockedStatus,
     WaitingStatus,
 )
 
-from service import SNAP_NAME, SnapResource, get_installed_snap_service, snap_install_or_refresh
+from service import (
+    SNAP_NAME,
+    SnapResource,
+    get_installed_snap_service,
+    snap_install_or_refresh,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -135,9 +139,10 @@ class OpenstackExporterOperatorCharm(ops.CharmBase):
         try:
             resource = SnapResource(RESOURCE_NAME, self.model)
             snap_install_or_refresh(resource, self.model.config["snap_channel"])
-        except SnapError:
+        except Exception as e:
+            logging.error("Failed to install/refresh openstack-exporter snap: %s", e)
             self.model.unit.status = BlockedStatus(
-                "Failed to remove/install openstack-exporter snap"
+                "Failed to install/refresh openstack-exporter snap"
             )
 
     def _configure(self, _: ops.HookEvent) -> None:
@@ -188,23 +193,32 @@ class OpenstackExporterOperatorCharm(ops.CharmBase):
         """Handle collect unit status event (called after every event)."""
         if not self.model.relations.get("credentials"):
             event.add_status(BlockedStatus("Keystone is not related"))
+            return
 
         if not self._get_keystone_data():
             event.add_status(WaitingStatus("Waiting for credentials from keystone"))
+            return
 
         if not self.model.relations.get("cos-agent"):
             event.add_status(BlockedStatus("Grafana Agent is not related"))
+            return
+
+        resource = SnapResource(RESOURCE_NAME, self.model)
+        if resource and resource.snap_name != SNAP_NAME:
+            event.add_status(BlockedStatus(f"snap resource {resource.snap_name} is not valid."))
+            return
 
         snap_service = get_installed_snap_service(SNAP_NAME)
-
         if not snap_service.present:
             event.add_status(
                 BlockedStatus("snap service is not installed, please check snap service")
             )
-        elif not snap_service.is_active():
+            return
+        if not snap_service.is_active():
             event.add_status(
                 BlockedStatus("snap service is not running, please check snap service")
             )
+            return
 
         event.add_status(ActiveStatus())
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -23,7 +23,7 @@ from ops.model import (
     WaitingStatus,
 )
 
-from service import SNAP_NAME, get_installed_snap_service, snap_install_or_refresh
+from service import SNAP_NAME, UPSTREAM_SNAP, get_installed_snap_service, snap_install_or_refresh
 
 logger = logging.getLogger(__name__)
 
@@ -168,8 +168,11 @@ class OpenstackExporterOperatorCharm(ops.CharmBase):
         """
         self.install()
 
-        snap_service = get_installed_snap_service(SNAP_NAME)
+        upstream_snap = get_installed_snap_service(UPSTREAM_SNAP)
+        if upstream_snap.present:
+            return
 
+        snap_service = get_installed_snap_service(SNAP_NAME)
         # if cos is not related then we should block and not run anything
         if not self.model.relations.get("cos-agent"):
             snap_service.stop()
@@ -213,6 +216,15 @@ class OpenstackExporterOperatorCharm(ops.CharmBase):
 
         if not self.model.relations.get("cos-agent"):
             event.add_status(BlockedStatus("Grafana Agent is not related"))
+
+        upstream_snap = get_installed_snap_service(UPSTREAM_SNAP)
+
+        if upstream_snap.present:
+            event.add_status(
+                BlockedStatus(
+                    f"{UPSTREAM_SNAP} should not be used anymore. Please remove the snap resource."
+                )
+            )
 
         snap_service = get_installed_snap_service(SNAP_NAME)
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -15,18 +15,14 @@ from typing import Any
 import ops
 import yaml
 from charms.grafana_agent.v0.cos_agent import COSAgentProvider
+from charms.operator_libs_linux.v2.snap import SnapError
 from ops.model import (
     ActiveStatus,
     BlockedStatus,
     WaitingStatus,
 )
 
-from service import (
-    SNAP_NAME,
-    SnapResource,
-    get_installed_snap_service,
-    snap_install_or_refresh,
-)
+from service import SNAP_NAME, SnapResource, get_installed_snap_service, snap_install_or_refresh
 
 logger = logging.getLogger(__name__)
 
@@ -139,10 +135,9 @@ class OpenstackExporterOperatorCharm(ops.CharmBase):
         try:
             resource = SnapResource(RESOURCE_NAME, self.model)
             snap_install_or_refresh(resource, self.model.config["snap_channel"])
-        except Exception as e:
-            logging.error("Failed to install/refresh openstack-exporter snap: %s", e)
+        except SnapError:
             self.model.unit.status = BlockedStatus(
-                "Failed to install/refresh openstack-exporter snap"
+                "Failed to remove/install openstack-exporter snap"
             )
 
     def _configure(self, _: ops.HookEvent) -> None:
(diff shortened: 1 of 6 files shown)


aieri:
I think I would prefer a less opaque behavior: asking the operator to explicitly remove golang-openstack-exporter installed as a resource. If we silently switch to charmed-openstack-exporter as fetched from charmhub it will be a lot less obvious by looking at the charm configuration to understand exactly what has been installed.

src/service.py
```diff
@@ -56,7 +61,52 @@ def present(self) -> bool:
...
-def snap_install_or_refresh(resource: Optional[str], channel: str) -> None:
+class SnapResource:
+    """A class representing the snap resource."""
+
+    def __init__(self, name: str, model: Model) -> None:
+        """Initialize the class."""
+        self.name = name
+        self.model = model
+
+    @property
```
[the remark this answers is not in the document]
gabrielcocenza: May I ask why?
I couldn't find a pep saying that this is a bad practice

src/service.py
```diff
@@ -4,11 +4,16 @@
 from logging import getLogger
 from typing import Any, Optional
 
+import yaml
 from charms.operator_libs_linux.v2 import snap
+from ops.model import Model, ModelError
+from PySquashfsImage import SquashFsImage
```
aieri, line 10: instead of using a 3rd party library I would prefer to use snap tooling for security reasons, even if we have to use subprocess:
```
$ snap download golang-openstack-exporter
Fetching snap "golang-openstack-exporter"
Fetching assertions for "golang-openstack-exporter"
Install the snap with:
   snap ack golang-openstack-exporter_97.assert
   snap install golang-openstack-exporter_97.snap

$ snap info ./golang-openstack-exporter_97.snap                        
path:       "golang-openstack-exporter_97.snap"
name:       golang-openstack-exporter
summary:    OpenStack Exporter for Prometheus
version:    v1.4.0 -
build-date: 2021-05-26
license:    unset
description: |
  The OpenStack exporter, exports Prometheus metrics from a running OpenStack
  cloud for consumption by prometheus. The cloud credentials and identity
  configuration should use the os-client-config format and must by specified
  with the --os-client-config flag.
  
  Other options as the binding address/port can by explored with the --help
  flag.
  
  By default the openstack_exporter serves on port 0.0.0.0:9180 at the /metrics
  URL.
commands:
  - golang-openstack-exporter.openstack-exporter
```

aieri (changes requested):

src/service.py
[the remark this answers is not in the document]
samuelallan72: There's no pep as far as I know, but `@property` violates the principle of least surprise (from the caller side it looks like an attribute, but it actually calls a method, which may have side effects or non-zero cost).  Fine to use it for certain cases, but in general there's no downside to explicitly calling a method.

src/service.py
samuelallan72: +1 to no extra dependencies; shelling out to `snap` is a nice solution :ok_hand:

gabrielcocenza: My initial implementation was using the subprocess and parsing the output, but I think this solution is very fragile. Any change in the output of the cli tool and this crashes the solution. They don't have any commitment on keeping the output as it is.

If we are going down that path I think I prefer using subprocess using `squashfs` and then reading the files

src/charm.py
```diff
@@ -207,23 +193,32 @@ def _on_collect_unit_status(self, event: ops.CollectStatusEvent) -> None:
         """Handle collect unit status event (called after every event)."""
         if not self.model.relations.get("credentials"):
             event.add_status(BlockedStatus("Keystone is not related"))
+            return
```
chanchiwai-ray, line 196: Are you sure you want to `return`? The status works like a "priority queue", you can add multiple statuses in one go, and the it will be evaluated and prioritize based on the importance mentioned [here](https://ops.readthedocs.io/en/latest/#ops.CollectStatusEvent). The final status is set based on the "priority queue" so you don't need to **explicitly** order the status in this handler.

chanchiwai-ray (commented):
Can't we simply keep the existing implementation but add an extra check to the snap? e.g. you can check if the snap is named `golang-openstack-exporter` or you can check if the snap has `service` (upstream snap does not have it because it's outdated). And based on the check, you can set an appropriate status

gabrielcocenza:
@chanchiwai-ray 

> Can't we simply keep the existing implementation but add an extra check to the snap? e.g. you can check if the snap is named golang-openstack-exporter.

This was a very nice suggestion, thanks!

I think it makes everything simpler.

@aieri @samuelallan72 I ended up applying Rays suggestion and then we don't need to check the resource snap name. Installing anyways and checking on config_changed and update_status makes things easier.

src/charm.py
gabrielcocenza: I didn't know about this. Thanks for the information

tests/integration/tests/charm_tests/openstack_exporter.py
```diff
@@ -256,18 +257,45 @@ def test_openstack_exporter_snap_down(self):
...
+
+            resource_file = tmp_path / "golang-openstack-exporter_amd64.snap"
+            model.attach_resource(APP_NAME, "openstack-exporter", str(resource_file))
+            model.block_until_unit_wl_status(
+                self.leader_unit_entity_id, "blocked", timeout=STATUS_TIMEOUT
+            )
+            snaps_installed = [snap["name"] for snap in get_snaps_installed()]
+            self.assertIn(UPSTREAM_SNAP, snaps_installed)
+            expected_msg = (
+                f"{UPSTREAM_SNAP} should not be used anymore. " "Please remove the snap resource."
```
samuelallan72, line 279:
```suggestion
                f"{UPSTREAM_SNAP} should not be used anymore. Please remove the snap resource."
```

tests/integration/tests/charm_tests/openstack_exporter.py
```diff
@@ -256,18 +257,45 @@ def test_openstack_exporter_snap_down(self):
...
+    def test_upstream_exporter_as_resource(self):
+        """Test using the upstream golang as resource will block the unit."""
+        # juju cannot access resources at /tmp, so we create a tmp folder at the current directory
+        with tempfile.TemporaryDirectory(dir="./") as tmpdir:
+            tmp_path = Path(tmpdir)
+            download_cmd = (
+                f"wget -q -P {tmpdir} https://github.com/canonical/openstack-exporter-operator/"
+                "releases/download/rev2/golang-openstack-exporter_amd64.snap"
+            )
+            subprocess.run(download_cmd.split(), check=False)
```
samuelallan72, line 269: I'm not sure about this... I don't think we can rely on an old release snap download url.  I guess it's fine to keep for now; eventually we can drop the logic that manages the old snap altogether after a certain deprecation period.

samuelallan72 (approved):
Thanks, this logic is nice - simple and appears to handle this case gracefully :+1:

tests/integration/tests/charm_tests/openstack_exporter.py
```diff
@@ -256,18 +257,45 @@ def test_openstack_exporter_snap_down(self):
...
-    snap_info = results.splitlines()[1].split()
-    return {
-        "version": snap_info[VERSION_COLUMN],
-        "revision": snap_info[REVISION_COLUMN],
-        "channel": snap_info[CHANNEL_COLUMN],
-    }
+def get_snap_info(snap_name: str) -> list[dict[str, str]]:
+    """Get the snap information."""
+    snaps = get_snaps_installed()
+    return [snap for snap in snaps if snap["name"] == snap_name][0]
```
chanchiwai-ray, line 301: I didn't know we can do that 👀. Maybe this is an easier and more efficient way of getting one snap info, e.g.

```
curl -sS --unix-socket /run/snapd.socket http://localhost/v2/find -G -d "name=juju"
```

Reference: https://github.com/canonical/operator-libs-linux/blob/20880b1dfba4f8e38a8026bb06ea9d3d30f6a081/lib/charms/operator_libs_linux/v2/snap.py#L785

chanchiwai-ray (approved):
Overall LGTM, one quality improvement (non-blocking)

tests/integration/tests/charm_tests/openstack_exporter.py
nobuto-m, line 279: By looking at this block message and the necessary step below, I believe it's worth an upgrade document. And the wording "please remove" may be misleading if Juju doesn't have a way to **remove** the resource. It would be good to suggest "by uploading an empty resource" or something like that.

tests/integration/tests/charm_tests/openstack_exporter.py
gabrielcocenza: Nice finding, I didn't know it either.
However I think this is not applicable to this case because I want the information from the snap installed and this gives the information of the snap from the snapstore

tests/integration/tests/charm_tests/openstack_exporter.py
gabrielcocenza: I've changed the wording to add an empty file and pointing also to the [known-issues](https://charmhub.io/openstack-exporter#known-issues) charmhub page.

We'll update the [tutorial](https://charmhub.io/openstack-exporter/docs/tutorial#openstack-exporter-snap) to not use the resource when this changes land into stable.

src/charm.py
```diff
@@ -214,6 +217,17 @@ def _on_collect_unit_status(self, event: ops.CollectStatusEvent) -> None:
...
 
+        upstream_snap = get_installed_snap_service(UPSTREAM_SNAP)
+
+        # this is necessary when doing a charm upgrade coming from revision 27
+        if upstream_snap.present:
+            event.add_status(
+                BlockedStatus(
+                    "Please add an empty file as resource. For more info see: "
+                    "https://charmhub.io/openstack-exporter#known-issues"
+                )
```
chanchiwai-ray, line 228: Juju's messages are usually clipped when it's too long, so I am not sure if the link will be shown in `juju status`

chanchiwai-ray (approved):
Some UX improvements, but it can be done later if it's turned out be an issue

tests/integration/tests/charm_tests/openstack_exporter.py
chanchiwai-ray: Thanks, I should double checked 🙂

src/charm.py
gabrielcocenza: Before it wasn't, but then I changed to fit.

tests/integration/tests/charm_tests/openstack_exporter.py
aieri: docs/tutorial updates are now live

src/charm.py
```diff
@@ -214,6 +217,17 @@ def _on_collect_unit_status(self, event: ops.CollectStatusEvent) -> None:
...
             event.add_status(BlockedStatus("Grafana Agent is not related"))
 
+        upstream_snap = get_installed_snap_service(UPSTREAM_SNAP)
+
+        # this is necessary when doing a charm upgrade coming from revision 27
+        if upstream_snap.present:
+            event.add_status(
+                BlockedStatus(
+                    "Please add an empty file as resource. For more info see: "
+                    "https://charmhub.io/openstack-exporter#known-issues"
```
nobuto-m, line 227: It's not a code issue but more of the doc issue outside of this repository but I comment about it here.

First of all, It's great to see the initiative to prepare upgrade notes for this kind of migration. Huge thanks for that.

The link states:
https://charmhub.io/openstack-exporter#known-issues
> Upgrading the charm from revision ≤ 27 to a revision higher than 28 may cause the snap to move a blocked state and require manual intervention. This is intentional.

It's a little bit awkward to have <= 27 and > 28 as the condition. I know the rev 28 was a bit problematic in this context but it won't hit to the stable channel and will be overridden by a new revision soon. So from a documentation perspective, it can be simplified as "when upgrading from <= 27" instead of mentioning to a target revision.

> Please see the related tutorial.

There is no link to the actual tutorial. Would be good to add a hyperlink to https://charmhub.io/openstack-exporter/docs/upgrade

aieri, line 227: We can link to the new tutorial here:
```suggestion
                    "Upgrades from revisions older than 28 require manual intervention. Please see: "
                    "https://charmhub.io/openstack-exporter/docs/upgrade-from-older-than-28"
```

tests/integration/tests/charm_tests/openstack_exporter.py
nobuto-m: I added a follow up comment https://github.com/canonical/openstack-exporter-operator/pull/93#discussion_r1737620856

aieri (changes requested):
small tweak proposed, otherwise lgtm

src/service.py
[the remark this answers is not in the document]
gabrielcocenza: Closing this once we are not having this class anymore.

src/service.py
gabrielcocenza: Closing this topic once we are not reading the name from the resource file anymore

src/charm.py
aieri: Right, mentioning r28 is a little awkward as it's not a stable release. I've tweaked the wording to use r27 as the pivot instead and fixed the hyperlinks (this change only affected 3 pages but made me feel very sorry for web developers that work on refactorings for big websites :rofl: )

aieri: if the original message fits but the proposed one does not I'm fine with keeping the one that fits

gabrielcocenza: It does not fit:
```
Model  Controller  Cloud/Region         Version  SLA          Timestamp
test   lxd         localhost/localhost  3.4.5    unsupported  18:05:28-03:00

App                    Version          Status   Scale  Charm               Channel        Rev  Exposed  Message
grafana-agent                           blocked      1  grafana-agent       latest/stable  216  no       Missing ['grafana-cloud-config']|['grafana-dashboards-provider']|['logging-consumer']|['send-remote-write'] for cos-a...
keystone               21.0.1           active       1  keystone            yoga/stable    689  no       Application Ready
keystone-mysql-router  8.0.39           active       1  mysql-router        8.0/stable     200  no       Unit is ready
mysql                  8.0.36-0ubun...  active       1  mysql               8.0/stable     240  no       
openstack-exporter                      blocked      1  openstack-exporter                  11  no       Upgrades from revisions older than 28 require manual intervention. Please see https://charmhub.io/openstack-exporter/...

Unit                        Workload  Agent  Machine  Public address  Ports           Message
keystone/0*                 active    idle   1        10.101.82.139   5000/tcp        Unit is ready
  keystone-mysql-router/0*  active    idle            10.101.82.139                   Unit is ready
mysql/0*                    active    idle   2        10.101.82.220   3306,33060/tcp  Primary
openstack-exporter/0*       blocked   idle   0        10.101.82.235                   Upgrades from revisions older than 28 require manual intervention. Please see https://charmhub.io/openstack-exporter/...
  grafana-agent/0*          blocked   idle            10.101.82.235                   Missing ['grafana-cloud-config']|['grafana-dashboards-provider']|['logging-consumer']|['send-remote-write'] for cos-a...

Machine  State    Address        Inst id        Base          AZ  Message
0        started  10.101.82.235  juju-966319-0  ubuntu@22.04      Running
1        started  10.101.82.139  juju-966319-1  ubuntu@22.04      Running
2        started  10.101.82.220  juju-966319-2  ubuntu@22.04      Running
```

aieri: How about
```suggestion
                    "golang-openstack-exporter detected. Please see: "
                    "https://charmhub.io/openstack-exporter#known-issues"
```

aieri (approved):

chanchiwai-ray (approved):

tests/integration/tests/charm_tests/openstack_exporter.py
gabrielcocenza: Yes, I agree that we can possibly remove this logic after some time.

src/charm.py
nobuto-m: I'm happy with the documentation updates. I don't have a privilege to mark the conversation as resolved since I'm not a collaborator. But feel free to close this.

PR #101, opened 2024-09-13
refactor: Address review comments on #100

jneo8:
- Remove `# TODO: can we remove the base64 encoding?` on `.github/workflows/check.yaml`
- Add explanation when both `openstack-exporter` and `snap_channel` are set.
- Change string-formatting to f-string
- Remove double negative
- Raise SnapError directly
- Add explanation why we need snap local installation


.github/workflows/check.yaml
```diff
@@ -72,7 +72,6 @@ jobs:
       - name: external controller juju config
         run: |
           mkdir -p $HOME/.local/share/juju
-          # TODO: can we remove the base64 encoding?
```
samuelallan72, line 75: I think we can add a comment about why it's base64 encoded.  I've gone down that rabbit hole before... perhaps in this case it's to avoid issues with templating a multiline string with possible quoting characters, etc. into a shell command?

.github/workflows/check.yaml
```diff
@@ -95,7 +94,7 @@ jobs:
 
       - name: install external controller required packages
         run: |
-           sudo snap install charmcraft --channel 2.x/stable --classic
```
samuelallan72, line 98: switching charmcraft channel is something we should do separately:

- update charmcraft 2 -> 3 everywhere in the repo (including the workflows)
- update charmcraft.yaml to follow latest conventions
- test it
- port changes to the workflows to the automation repo

charmcraft.yaml
```diff
@@ -89,3 +91,5 @@ resources:
       scraping. The snap resource on Charmhub is intended to be an empty file. The charm will
       ignore an empty file resource and normally install the snap from the snap store, but a
       custom snap can also be provided here if needed.
+
+      This resource has top priority when the install hook is triggered. The snap_channel option will have no effect.
```
samuelallan72, line 95: maybe for clarity, something like this?

```suggestion
      This resource has top priority. If the resource is present, the snap_channel option will have no effect.
```

(this is user facing documentation, so mentioning install hook triggered is implementation details they don't need to know)

samuelallan72 (changes requested):

.github/workflows/check.yaml
jneo8: Ok, let's do it separately. I will remove this change.

.github/workflows/check.yaml
jneo8: Updated

.github/workflows/check.yaml
jneo8: Deleted, so resolved.

charmcraft.yaml
jneo8: Updated

src/service.py
```diff
@@ -47,7 +49,7 @@ def configure(self, snap_config: dict[str, Any]) -> None:
         # Bail out or it will crash on self.snap_client.set() method
         if not snap_config:
             logger.warning("empty snap config: %s, skipping...", snap_config)
-            return
+            raise SnapConfigurationMissingError
```
samuelallan72, line 52: Is this caught anywhere?  How is this better than `snap_client.set()` crashing?

samuelallan72 (dismissed):

src/charm.py
```diff
@@ -147,7 +140,7 @@ def get_resource(self) -> Optional[str]:
             logger.debug("cannot fetch charm resource")
             return None
 
-        if not os.path.getsize(snap_path) > 0:
+        if not os.path.getsize(snap_path):
```
chanchiwai-ray, line 143: This is logically weird to understand, although it has the same effect. `os.path.getsize` return number, and `not <number>` is not logic (although it makes sense in Python)

src/service.py
chanchiwai-ray: If we really want to raise exception,can we use the one raised by `snap`?

```
self.snap_client.set(snap_config, typed=True)
```

jneo8: Usually I prefer self-defined error because the message or naming is more easy to understand. I mean if you compare `SnapError` vs `SnapConfigurationMissingError`, `SnapConfigurationMissingError` give you more information.

chanchiwai-ray: How's that give you more information, while the error raised by `snap` provide the reason why it failed e.g.`snap.SnapError: Snap: 'rustup'; command ['snap', 'set', 'rustup', '-t'] failed with output = ''
`

src/charm.py
jneo8: Or we can change it to:

```python
# No double negative, and size > 0 return boolean.
if os.path.getsize(snap_path) > 0:
    return snap_path
else:
    logger.debug("resource is an empty file")
    return None
```

(Or any recommend?)

samuelallan72: I'm happy either way.  I do agree that retaining a comparison with an integer makes it easier to read the condition.  A third option if we don't want to reorder the code and want to avoid the double negative:

```python
if os.path.getsize(snap_path) == 0:
    .... return None

...
```

jneo8: I change it to

```
if os.path.getsize(snap_path) <= 0
```

to avoid double negative.

src/service.py
chanchiwai-ray: In addition to this, if we raise here, we should also catch in `charm.py` as @samuelallan72 mentioned

jneo8: It's different, one is empty, another is wrong configuration.

chanchiwai-ray: > It's different, one is empty, another is wrong configuration.

That's the internal error (actual error) because `snap_client.set` simply run `subprocess` under the hook. So if you pass a `{}` to `snap_client.set`  you are actually running `snap set openstack-exporter -t `, and that's missing an argument

jneo8: Updated

samuelallan72 (approved):

chanchiwai-ray (dismissed):

chanchiwai-ray (approved):

PR #109, opened 2024-09-24
Remove duplicate application from func test bundle

samuelallan72:


tests/integration/tests/bundles/jammy-yoga.yaml
```diff
@@ -93,9 +93,6 @@ applications:
       openstack-origin: distro
       verbose: true
     constraints: arch=amd64 mem=2048
-  nova-cloud-controller-mysql-router:
```
rgildein, line 96: There are still relations for this app.

rgildein (changes requested):

tests/integration/tests/bundles/jammy-yoga.yaml
chanchiwai-ray: It's duplicated, see line 83. However, it's weird that juju accept this bundle file

jneo8 (approved):

rgildein (approved):
LGTM

gabrielcocenza (approved):

Issue #108, opened 2024-09-20
Should be in error state if snap service down and uninstallation failed


Pjack:
Address the issue mentioned in 

* Snap service is down
  * https://github.com/canonical/openstack-exporter-operator/pull/100#discussion_r1758186715

* Un-installation failed
  * https://github.com/canonical/openstack-exporter-operator/pull/100#discussion_r1758175152
  * https://github.com/canonical/openstack-exporter-operator/pull/100#discussion_r1758179346

* Snap installtion failed (I found it additionally)
  * https://github.com/canonical/openstack-exporter-operator/blob/main/src/charm.py#L154

Pjack:
After discussing with @jdkandersson, I believe we can adopt this practice in the GitHub runner. 
We don't set the charm to ErrorStatus; instead, we allow the exceptions to bubble up and be handled by the Ops framework.
https://github.com/canonical/github-runner-operator/blob/main/src/charm.py#L454

That's to say. If an unexpected error occurs in the workload (e.g., installation failure, service downtime), raise an exception in the charm if no specific action can be provided to resolve the issue. After the operator reviews the workload logs, they can manually resolve Juju. 

What do you think ?

jneo8:
+1 for error status. It's a diverge situation so we can't provide a document or instruction to the user.

PR #110, opened 2024-09-24
Raise errors on unrecoverable problems

samuelallan72:
Raise errors for snap install or refresh failure.  Raise an error because it's a fatal failure, and juju will auto-retry the hook, which may succeed on retry if the error was transient.

Improve blocked status message for if it detects snap not installed or service not active in the status hook.
I think Blocked is better than Error status here:
- auto retrying the status hook isn't going to fix anything
- blocked status lets the charm provide a more useful status message to the user
- blocked status will automatically update to active on the next update-status hook once the issue is resolved

Partially fixes: #108

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -16,7 +16,6 @@ from typing import Any, Optional
 import ops
 import yaml
 from charms.grafana_agent.v0.cos_agent import COSAgentProvider
-from charms.operator_libs_linux.v2.snap import SnapError
 from ops.model import ActiveStatus, BlockedStatus, ModelError, WaitingStatus
 
 from service import SNAP_NAME, UPSTREAM_SNAP, get_installed_snap_service, snap_install_or_refresh
--- a/src/charm.py
+++ b/src/charm.py
@@ -149,6 +149,7 @@ class OpenstackExporterOperatorCharm(ops.CharmBase):
         """Install the necessary resources for the charm."""
         # If this fails, it's not recoverable.
         # So we don't catch the error, instead letting this become a charm error status.
+        # Errored hooks are auto-retried by juju, so the install may work on retry.
         snap_install_or_refresh(self.get_resource(), self.model.config["snap_channel"])
 
     def _configure(self, _: ops.HookEvent) -> None:
--- a/src/charm.py
+++ b/src/charm.py
@@ -148,12 +148,9 @@ class OpenstackExporterOperatorCharm(ops.CharmBase):
 
     def install(self) -> None:
         """Install the necessary resources for the charm."""
-        try:
-            snap_install_or_refresh(self.get_resource(), self.model.config["snap_channel"])
-        except SnapError:
-            self.model.unit.status = BlockedStatus(
-                "Failed to remove/install openstack-exporter snap"
-            )
+        # If this fails, it's not recoverable.
+        # So we don't catch the error, instead letting this become a charm error status.
+        snap_install_or_refresh(self.get_resource(), self.model.config["snap_channel"])
 
     def _configure(self, _: ops.HookEvent) -> None:
         """Configure the charm.
@@ -227,13 +224,9 @@ class OpenstackExporterOperatorCharm(ops.CharmBase):
         snap_service = get_installed_snap_service(SNAP_NAME)
 
         if not snap_service.present:
-            event.add_status(
-                BlockedStatus("snap service is not installed, please check snap service")
-            )
+            raise RuntimeError(f"{SNAP_NAME} snap is not installed")
         elif not snap_service.is_active():
-            event.add_status(
-                BlockedStatus("snap service is not running, please check snap service")
-            )
+            raise RuntimeError(f"{SNAP_NAME} snap service is not active")
 
         event.add_status(ActiveStatus())
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -51,8 +51,10 @@ class OpenstackExporterOperatorCharm(ops.CharmBase):
         self.framework.observe(self.on.upgrade_charm, self._on_upgrade)
         self.framework.observe(self.on.config_changed, self._configure)
         self.framework.observe(self.on.collect_unit_status, self._on_collect_unit_status)
+        self.framework.observe(self.on.credentials_relation_joined, self._configure)
         self.framework.observe(self.on.credentials_relation_changed, self._configure)
         self.framework.observe(self.on.credentials_relation_broken, self._configure)
+        self.framework.observe(self.on.cos_agent_relation_joined, self._configure)
         self.framework.observe(self.on.cos_agent_relation_changed, self._configure)
         self.framework.observe(self.on.cos_agent_relation_broken, self._configure)
 
@@ -203,16 +205,18 @@ class OpenstackExporterOperatorCharm(ops.CharmBase):
         """Handle collect unit status event (called after every event)."""
         if not self.model.relations.get("credentials"):
             event.add_status(BlockedStatus("Keystone is not related"))
+            return
 
         if not self._get_keystone_data():
             event.add_status(WaitingStatus("Waiting for credentials from keystone"))
+            return
 
         if not self.model.relations.get("cos-agent"):
             event.add_status(BlockedStatus("Grafana Agent is not related"))
-
-        upstream_snap = get_installed_snap_service(UPSTREAM_SNAP)
+            return
 
         # this is necessary when doing a charm upgrade coming from revision 27
+        upstream_snap = get_installed_snap_service(UPSTREAM_SNAP)
         if upstream_snap.present:
             event.add_status(
                 BlockedStatus(
@@ -220,17 +224,15 @@ class OpenstackExporterOperatorCharm(ops.CharmBase):
                     "https://charmhub.io/openstack-exporter#known-issues"
                 )
             )
+            return
 
         snap_service = get_installed_snap_service(SNAP_NAME)
 
         if not snap_service.present:
-            event.add_status(
-                BlockedStatus(f"{SNAP_NAME} snap is not installed; please manually check")
-            )
-        elif not snap_service.is_active():
-            event.add_status(
-                BlockedStatus(f"{SNAP_NAME} snap service is not active; please manually check")
-            )
+            raise RuntimeError(f"{SNAP_NAME} snap is not installed, but it should be")
+
+        if not snap_service.is_active():
+            raise RuntimeError(f"{SNAP_NAME} snap service is not active, but it should be")
 
         event.add_status(ActiveStatus())
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -225,11 +225,19 @@ class OpenstackExporterOperatorCharm(ops.CharmBase):
 
         if not snap_service.present:
             event.add_status(
-                BlockedStatus(f"{SNAP_NAME} snap is not installed. Please wait for installation to complete, or manually reinstall the snap if the issue persists.")
+                BlockedStatus(
+                    f"{SNAP_NAME} snap is not installed. "
+                    "Please wait for installation to complete, "
+                    "or manually reinstall the snap if the issue persists."
+                )
             )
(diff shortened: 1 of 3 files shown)


rgildein (changes requested):
I like this change, since it make more sense than block status.
Please fix lint a and unit tests. Can we also add a simple integration test for this?

src/charm.py
```diff
@@ -227,13 +224,9 @@ def _on_collect_unit_status(self, event: ops.CollectStatusEvent) -> None:
...
         if not snap_service.present:
-            event.add_status(
-                BlockedStatus("snap service is not installed, please check snap service")
-            )
+            raise RuntimeError(f"{SNAP_NAME} snap is not installed")
         elif not snap_service.is_active():
-            event.add_status(
-                BlockedStatus("snap service is not running, please check snap service")
-            )
+            raise RuntimeError(f"{SNAP_NAME} snap service is not active")
```
chanchiwai-ray, line 229: I am concerned this will not work well with `_on_collect_unit_status`

chanchiwai-ray (changes requested):
If we insist on raising error, I recommend we rework the `_on_collect_unit_status`

src/charm.py
jneo8: I think `event.add_status` can work with ErrorStatus.

See: https://github.com/canonical/operator/blob/98a4e4f44d37d3361a253bc1eff49eee2ca6d140/ops/model.py#L1960-L1975

chanchiwai-ray: (not verified) Still, it should not work because `ErrorStatus` has the highest priority. It means that install event -> collect status -> raise Error status. This will happen whether or not you remove the cos-agent relation check

samuelallan72:
Thanks for reviews everyone - I see this is early stage still, I probably could have left it as a draft.

src/charm.py
samuelallan72: I've been recommended before never to manually set ErrorStatus - it's reserved for actual errors.  See also https://github.com/canonical/operator/blob/9664fd1ad43992e0aa3fb3ede9213f886eba61ac/ops/model.py#L1834-L1835 that suggests ops framework will block it if you manually set errorstatus.

samuelallan72: > I am concerned this will not work well with _on_collect_unit_status.

I wasn't sure how to handle this.  There are two main paths this condition could be hit in collect status:

1. the original install or server start failed - that should completely shadow anything set from collect status though, because errorstatus is higher priority, and besides it shouldn't get to the update-status hook unless the user cancels retrying the install or configure hooks.  In this case, error seems fine to raise, because we've got that unrecoverable condition.
2. something changed on the unit (a user or something else caused the snap to get uninstalled, or the service disabled, or something is crashing the service, etc.) - here is where update-status is the only place to get a timely warning about that.  Again error is probably most appropriate given the feedback we received from the charm listing review.

I can't find docs, but I guess if ops framework gracefully handles raising an error, it should work the same in collect_status?

samuelallan72: > Edit: and it's true, the charm failed on install hook immediately because it's missing cos-agent relation

Ah I just tested, and now I understand your comment.  We can't unconditionally assert that the service is active, because the charm doesn't have the service running unless cos-agent and keystone are ready.

chanchiwai-ray: I think even if you remove "cos-agent and keystone are ready" check and start the exporter during config changed event, it still won't work. Because the collect status is run after install hook: install -> collect status (crash here) -> config changed -> collect status -> ...

samuelallan72: Updated - I went the other way - only check if the snap service is running if the charm expects it to be running (ie. keystone data is ready and cos-agent is related).

samuelallan72:
Ok lints and unit tests fixed, ready for review again.

A downside I see of raising errors instead of using a custom blocked status, is that the juju status message is not as useful:

```
openstack-exporter/0*  error     idle   0        10.168.106.56             hook failed: "update-status"
```

I'm not sure if we can improve that while still following this pattern of raising the error and letting the ops framework manage the error status.

samuelallan72:
Actually I think this needs more work on the conditions to check if the exporter is running.  I just observed this:

```
openstack-exporter/0*  error     idle   1        10.168.106.220            hook failed: "credentials-relation-joined"
```

I guess this is because configure is not called for credentials-relation-joined hook, so at the end of the credentials-relation-joined hook, collect_status is run, and at this point the charm has the keystone data, but the service hasn't been started yet.  The service will start in the next credentials-relation-configure hook.

samuelallan72:
I updated it to avoid false positives with raising the error:

- ensure that configuring is run for the join hooks too, so collect_status can always pick up a consistent state (otherwise it could detect that the exporter service should be running based on relations joined and data available, but in fact the charm hasn't started it yet).
- add early returns to collect_status to avoid adding more logic

src/charm.py
chanchiwai-ray: ha, @jneo8 they just updated `ops` days ago so that adding `ErrorStatus` or `UnknownStatus` will crash the charm

src/charm.py
```diff
@@ -206,34 +205,34 @@ def _on_collect_unit_status(self, event: ops.CollectStatusEvent) -> None:
...
 
         if not self._get_keystone_data():
             event.add_status(WaitingStatus("Waiting for credentials from keystone"))
+            return
 
         if not self.model.relations.get("cos-agent"):
             event.add_status(BlockedStatus("Grafana Agent is not related"))
-
-        upstream_snap = get_installed_snap_service(UPSTREAM_SNAP)
+            return
```
chanchiwai-ray, line 216: Interesting, search all `event.add_status` in canonical orgs some are using `event.add_status` follow by `return` some don't. I don't quite get the idea of `return` after `event.add_status`, it makes the priority feature useless.

samuelallan72: Priority feature is useful, however here we want to shortcircuit the logic.  Otherwise we need to re-check the conditions above before checking for snap_service.is_active()

chanchiwai-ray: https://github.com/canonical/sdcore-upf-operator/blob/b04abd2e64a17daabc043b5e2b131a889ddeeb87/src/charm.py#L72 vs
https://github.com/canonical/identity-platform-admin-ui-operator/blob/d109f04511c171b5c4f65bb98d170da569e7039d/src/charm.py#L321

samuelallan72:
Not sure what's going on with the integration tests. :/

src/charm.py
Pjack: I prefer short-circuiting; the other approach smells full of potential and unexpected bug.

src/charm.py
```diff
@@ -206,34 +205,34 @@ def _on_collect_unit_status(self, event: ops.CollectStatusEvent) -> None:
...
 
         if not snap_service.present:
-            event.add_status(
-                BlockedStatus("snap service is not installed, please check snap service")
-            )
-        elif not snap_service.is_active():
-            event.add_status(
-                BlockedStatus("snap service is not running, please check snap service")
-            )
+            raise RuntimeError(f"{SNAP_NAME} snap is not installed, but it should be")
```
Pjack, line 232: Will we see this message when we type `juju status`?

samuelallan72: Unfortunately not - it's a limitation of juju that it's not possible to set a juju status for error state.  It will look something like this:

```
openstack-exporter/0*  error     idle   0        10.168.106.56             hook failed: "update-status"
```

samuelallan72: I think I'm going back and forth on blocked vs error status here... I'm not sure what is best. :/

Pjack: I have two assumptions:

1. The configuration event handler won’t take effect until the installation is successful.
2. We can provide guidance to the user based on the most common scenario

If we've already handled the installation procedure in snap_install_or_refresh(), why do we need to check these two cases here? Is it possible that this handler could be triggered before the installation is complete? If so, would it be reasonable to provide general guidance to help users resolve the issue?

For example: 'Snap service is unavailable. Please wait for the installation to complete or manually reinstall the Smartctl snap if the issue persists.'

src/charm.py
```diff
@@ -206,34 +205,34 @@ def _on_collect_unit_status(self, event: ops.CollectStatusEvent) -> None:
...
+
+        if not snap_service.is_active():
+            raise RuntimeError(f"{SNAP_NAME} snap service is not active, but it should be")
```
Pjack, line 235: Ditto

For example: 'Snap service is not active. Please wait for the installation to complete or manually enable the Smartctl snap if the issue persists.'

src/charm.py
samuelallan72: > If we've already handled the installation procedure in snap_install_or_refresh(), why do we need to check these two cases here?

I guess it's to check if something has broken the exporter - like maybe the exporter crashed, or a colocated application/charm interfered, etc.

I like the idea of a more helpful message in juju status.

samuelallan72: I guess the point of concern is that if the snap service stops running or the snap is uninstalled, it's likely to be a deeper issue that couldn't be resolved simply by reinstalling or restarting manually. :thinking:

samuelallan72:
Integration tests are failing because the upgrade hook is crashing on collect status `RuntimeError: charmed-openstack-exporter snap service is not active, but it should be`.

It's making too many false positives for error state here.  If it was blocked, then this would result in a transient blocked state while the exporter wasn't running during an upgrade.

src/charm.py
samuelallan72: Updated

src/charm.py
samuelallan72: Updated :)

chanchiwai-ray (dismissed):

src/charm.py
Pjack: In the document 
https://ops.readthedocs.io/en/latest/#ops.CollectStatusEvent

> It is an error to call [add_status()](https://ops.readthedocs.io/en/latest/#ops.CollectStatusEvent.add_status) with an instance of [ErrorStatus](https://ops.readthedocs.io/en/latest/#ops.ErrorStatus) or [UnknownStatus](https://ops.readthedocs.io/en/latest/#ops.UnknownStatus).

src/charm.py
```diff
@@ -228,11 +225,19 @@ def _on_collect_unit_status(self, event: ops.CollectStatusEvent) -> None:
 
         if not snap_service.present:
             event.add_status(
-                BlockedStatus("snap service is not installed, please check snap service")
+                BlockedStatus(
+                    f"{SNAP_NAME} snap is not installed. "
+                    "Please wait for installation to complete, "
+                    "or manually reinstall the snap if the issue persists."
+                )
```
Pjack, line 232: I'm a bit confused now. Don't we need to add `return` right after "add_status"?

I initially thought the framework would select the highest-priority status across all units. However, for an individual unit, it seems more logical to return only one specific status. Returning blocked + active statuses at the same time feels odd to me. I'm unsure what the correct approach is, but I guess the final outcome would be the same.  

Here is the example code in document. 
https://ops.readthedocs.io/en/latest/#ops.CollectStatusEvent
```
    def _on_collect_status(self, event: ops.CollectStatusEvent):
        if 'port' not in self.model.config:
            event.add_status(ops.BlockedStatus('please set "port" config'))
            return
        event.add_status(ops.ActiveStatus())
```

Please feel free to close this conversation if the final outcome is the same. I just want to double check . thanks!

Pjack (approved):

src/charm.py
```diff
@@ -228,11 +225,19 @@ def _on_collect_unit_status(self, event: ops.CollectStatusEvent) -> None:
...
             )
         elif not snap_service.is_active():
             event.add_status(
-                BlockedStatus("snap service is not running, please check snap service")
+                BlockedStatus(
```
jneo8, line 236: **question (non-blocking):** 

Can you provide the explanation how this BlockStatus can resolved by charm automatically on the description?

samuelallan72: Sure, done. :)

jneo8 (approved):

Issue #113, opened 2024-10-18
snap_channel should be empty if resource attached


Pjack:
Follow up this suggestion in https://github.com/canonical/openstack-exporter-operator/pull/100#discussion_r1791861071

PR #115, opened 2024-10-20
Enter blocked status if resource and snap_channel

samuelallan72:
When a resource is used, the `snap_channel` config option is ignored. In this case, the snap is installed directly from the resource file, so a "channel" doesn't make sense.

If `snap_channel` is set and a resource is being used, it may indicate a confusion situation,
such as the user changing the `snap_channel`
and wondering why the snap isn't being updated.
So the charm can alert the user to this situation
via a message and blocked status.

Fixes: #113

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -228,8 +228,8 @@ class OpenstackExporterOperatorCharm(ops.CharmBase):
         if self.model.config["snap_channel"] != "latest/stable" and self.get_resource():
             event.add_status(
                 BlockedStatus(
-                    "The snap_channel option is unused when snap resource is provided. "
-                    f"Please unset it: juju config {self.app.name} --unset snap_channel"
+                    "snap_channel is unused when snap resource provided; "
+                    f"please unset it: juju config {self.app.name} --reset snap_channel"
                 )
             )
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -228,7 +228,8 @@ class OpenstackExporterOperatorCharm(ops.CharmBase):
         if self.model.config["snap_channel"] != "latest/stable" and self.get_resource():
             event.add_status(
                 BlockedStatus(
-                    "Unset snap_channel option: it is unused when snap resource is provided."
+                    "The snap_channel option is unused when snap resource is provided. "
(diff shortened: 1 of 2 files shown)


jneo8 (dismissed):
LGTM

src/charm.py
```diff
@@ -223,6 +223,15 @@ def _on_collect_unit_status(self, event: ops.CollectStatusEvent) -> None:
 
         snap_service = get_installed_snap_service(SNAP_NAME)
 
+        # The snap_channel config option is ignored if there is a snap resource,
+        # so warn the user.
+        if self.model.config["snap_channel"] != "latest/stable" and self.get_resource():
```
chanchiwai-ray, line 228: Why do we only care about the magic string `latest/stable`?

chanchiwai-ray (commented):
Overall LGTM, only one small question

src/charm.py
samuelallan72: Because `latest/stable` is the default value for snap_channel, and we don't have a generic way with the ops framework (as far as I know) to tell if the config has been user-set.

chanchiwai-ray: I am assuming it's related to the unset `snap_channel` option

chanchiwai-ray: > Because latest/stable is the default value for snap_channel, and we don't have a generic way with the ops framework (as far as I know) to tell if the config has been user-set.

Can we put the exact command for in the message? I guess some impatient user (like me) may accidentally use `juju config snap_channel=""`  when I see `unset`

samuelallan72: Good point, let me update it. :+1:

gabrielcocenza (approved):

src/charm.py
samuelallan72: Updated, tested, and ready for review again. :)

jneo8 (approved):

src/charm.py
chanchiwai-ray: thanks

chanchiwai-ray (approved):
LGTM

PR #116, opened 2024-10-24
Update to charmcraft 3

samuelallan72:
Fixes: #95

### DIFF
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -96,7 +96,7 @@ jobs:
 
       - name: install external controller required packages
         run: |
-           sudo snap install charmcraft --channel 2.x/stable --classic
+           sudo snap install charmcraft --channel 3.x/stable --classic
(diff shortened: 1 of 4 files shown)


chanchiwai-ray (approved):
LGTM, side question: do we need to update `charmcraft.yaml`?

samuelallan72:
> do we need to update charmcraft.yaml?

I believe not - it builds fine. :)

aieri (approved):

gabrielcocenza:
> LGTM, side question: do we need to update charmcraft.yaml?

I thought the same....

I imagine that it's not every project that will work

samuelallan72:
> I imagine that it's not every project that will work

Yes, some projects will not work, but most should be fine. We originally pinned everything to charmcraft 2 even if it seemed to work, because of some bugs in earlier releases of charmcraft 3 around handling of bases.

PR #127, opened 2024-12-12
Switch to private endpoint runners

samuelallan72:


samuelallan72:
The job runs on the new runner, but looks like it can't reach our existing controller. So I'll leave this for now, and resume dev when we have the new controller.

samuelallan72:
Closing because we're not going this route in the end.

PR #129, opened 2025-02-13
Run func tests locally on runner

samuelallan72:
Remove the dependency on the external juju controller, switching to running them locally on a larger github runner instance. In the process, update to the latest template from our automation.


chanchiwai-ray (approved):
Looks nice! Thanks

Pjack (approved):

jneo8 (approved):

PR #132, opened 2025-02-24
Support charmcraft 3 and multi platform builds

chanchiwai-ray:


samuelallan72 (approved):

charmcraft.yaml
```diff
@@ -8,13 +8,13 @@ description: |
...
-  - build-on:
-    - name: ubuntu
-      channel: "22.04"
-    run-on:
-    - name: ubuntu
-      channel: "22.04"
+parts:
+  charm:
+    plugin: charm
+    source: .
```
samuelallan72, line 14: I'm curious about why this was required? Strange that it worked without a parts definition at all... perhaps that's just the default definition?

chanchiwai-ray:
```
Bad charmcraft.yaml content:
- dictionary should have at least 1 item after validation, not 0 (in field 'parts')
```

Deezzir (approved):

charmcraft.yaml
```diff
@@ -8,13 +8,13 @@ description: |
...
+
+platforms:
+  ubuntu@22.04:amd64:
```
gabrielcocenza, line 17: Is there a reason why we are not adding ubuntu@20.04:amd64 and ubuntu@24.04:amd64 ?

samuelallan72: This charm has only supported jammy from release. We could probably add noble support, but I'm not sure if we'd hit issues with dependencies - we can maybe do that in a new ticket?

chanchiwai-ray: Yes, it wasn't there before

gabrielcocenza: Ok. Thanks for the info

PR #133, opened 2025-02-28
Attempt to fix functional tests

samuelallan72:
Functional tests are failing on main,
with errors from the mysql cluster.
It's potentially due to the model being too large to deploy on the runner and be stable, or it could be related to flakiness in the mysql-innodb-cluster charm. Either way, let's try to switch to the new mysql charm that can be deployed in non HA mode by only deploying a single unit. This should reduce resource use, and also perhaps be more stable.


samuelallan72:
Coming back to this - it appears func tests are passing on main again, so I'll close this.

PR #142, opened 2025-03-14
Soleng 576 Improve test coverage to 100%

### LINKED ISSUE #92: Project is not setting 100% coverage
This project doesn't set the tool coverage to 100% which can decrease the code quality.

Right now the project is at 66% coverage
```
unit run-test: commands[1] | coverage report
Name             Stmts   Miss  Cover   Missing
----------------------------------------------
src/charm.py        99     35    65%   72, 97-124, 131-136, 151-155, 175-177, 199-201, 213, 217-237
src/service.py      69     22    68%   23, 36-38, 42, 47-51, 121-131, 139-143
----------------------------------------------
TOTAL              168     57    66%
Coverage failure: total of 66 is less than fail-under=100
```

We need to reach to 100% and then change the configuration on `pyproject.toml`. Would also be nice to have a standard `pyproject` file into soleng automation to force new projects

H-M-Quang-Ngo:
Closes #92 
- Add more unit tests to reach 100% coverage
- Set coverage threshold to 100%

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,6 +5,7 @@ source = ["."]
 omit = ["tests/**", "docs/**", "lib/**", "snap/**", "build/**", "setup.py"]
 
 [tool.coverage.report]
+fail_under = 80
 show_missing = true
 
 [tool.coverage.html]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ source = ["."]
 omit = ["tests/**", "docs/**", "lib/**", "snap/**", "build/**", "setup.py"]
 
 [tool.coverage.report]
-fail_under = 80
+fail_under = 100
 show_missing = true
 
 [tool.coverage.html]
--- a/tests/unit/test_charm.py
+++ b/tests/unit/test_charm.py
@@ -109,118 +109,193 @@ class TestCharm:
         with pytest.raises(SnapError, match="My Error"):
             self.harness.charm.on.install.emit()
 
-    @mock.patch("charm.get_installed_snap_service")
-    @mock.patch("charm.OpenstackExporterOperatorCharm._get_keystone_data")
-    @mock.patch("charm.OpenstackExporterOperatorCharm.get_resource")
-    @mock.patch("charm.snap_install_or_refresh")
-    def test_on_collect_unit_status(
-        self,
-        _,
-        mock_get_resource,
-        mock_get_keystone_data,
-        mock_get_installed_snap_service,
-    ):
+    @pytest.mark.parametrize(
+        "config,relations,test_values,expected_status",
+        [
+            # Scenario 1: No credentials relation
+            (
+                {},  # Default config
+                [],  # No relations
+                {
+                    "keystone_data": {},
+                    "upstream_present": False,
+                    "snap_present": True,
+                    "snap_active": True,
+                    "resource_path": "/path/to/snap/resource",
+                },
+                ops.BlockedStatus("Keystone is not related"),
+            ),
+            # Scenario 2: With credentials relation but no data
+            (
+                {},  # Default config
+                [("credentials", "keystone")],
+                {
+                    "keystone_data": {},
+                    "upstream_present": False,
+                    "snap_present": True,
+                    "snap_active": True,
+                    "resource_path": "/path/to/snap/resource",
+                },
+                ops.WaitingStatus("Waiting for credentials from keystone"),
+            ),
+            # Scenario 3: No cos-agent relation
+            (
+                {},  # Default config
+                [("credentials", "keystone")],  # Only credentials relation
+                {
+                    "keystone_data": {"random": "data"},
+                    "upstream_present": False,
+                    "snap_present": True,
+                    "snap_active": True,
+                    "resource_path": "/path/to/snap/resource",
+                },
+                ops.BlockedStatus("Grafana Agent is not related"),
+            ),
+            # Scenario 4: With upstream snap present
+            (
+                {},  # Default config
+                [("credentials", "keystone"), ("cos-agent", "grafana-agent")],
+                {
+                    "keystone_data": {"random": "data"},
+                    "upstream_present": True,  # Upstream snap present
+                    "snap_present": True,
+                    "snap_active": True,
+                    "resource_path": "/path/to/snap/resource",
+                },
+                ops.BlockedStatus(
+                    "golang-openstack-exporter detected. Please see: "
+                    "https://charmhub.io/openstack-exporter#known-issues"
+                ),
+            ),
+            # Scenario 5: Non-default snap channel with resource
+            (
+                {"snap_channel": "latest/edge"},
+                [("credentials", "keystone"), ("cos-agent", "grafana-agent")],
+                {
+                    "keystone_data": {"random": "data"},
+                    "upstream_present": False,
+                    "snap_present": True,
+                    "snap_active": True,
+                    "resource_path": "/path/to/snap/resource",
+                },
+                ops.BlockedStatus(
+                    "Snap resource provided, so snap_channel is unused. "
+                    "Please unset it: juju config {app_name} --reset snap_channel"
+                ),
+            ),
+            # Scenario 6: Snap not installed
+            (
+                {},  # Default config
+                [("credentials", "keystone"), ("cos-agent", "grafana-agent")],
+                {
+                    "keystone_data": {"random": "data"},
+                    "upstream_present": False,
+                    "snap_present": False,
+                    "snap_active": False,
+                    "resource_path": "/path/to/snap/resource",
+                },
+                ops.BlockedStatus(
+                    f"{SNAP_NAME} snap is not installed. "
+                    "Please wait for installation to complete, "
+                    "or manually reinstall the snap if the issue persists."
+                ),
+            ),
+            # Scenario 7: Snap installed but service not active
+            (
+                {},  # Default config
+                [("credentials", "keystone"), ("cos-agent", "grafana-agent")],
+                {
+                    "keystone_data": {"random": "data"},
+                    "upstream_present": False,
+                    "snap_present": True,
+                    "snap_active": False,
+                    "resource_path": "/path/to/snap/resource",
+                },
+                ops.BlockedStatus(
+                    f"{SNAP_NAME} snap service is not active. "
+                    "Please wait for configuration to complete, "
+                    "or manually start the service the issue persists."
+                ),
+            ),
+            # Scenario 8: Everything ok
+            (
+                {"snap_channel": "latest/stable"},  # Default snap channel
+                [("credentials", "keystone"), ("cos-agent", "grafana-agent")],
+                {
+                    "keystone_data": {"random": "data"},
+                    "upstream_present": False,
+                    "snap_present": True,
+                    "snap_active": True,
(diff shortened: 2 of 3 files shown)


tests/unit/test_charm.py
```diff
@@ -107,3 +108,340 @@ def test_install_snap_error(self, mock_install):
...
+        # Scenario 1: No credentials relation
+        self.harness.charm._on_collect_unit_status(mock_event)
+        mock_event.add_status.assert_any_call(ops.BlockedStatus("Keystone is not related"))
+
+        # Scenario 2: With credentials relation but no data
+        self.harness.add_relation("credentials", "keystone")
+        self.harness.charm._on_collect_unit_status(mock_event)
+        mock_event.add_status.assert_any_call(
+            ops.WaitingStatus("Waiting for credentials from keystone")
+        )
```
gabrielcocenza, line 150: each unit tests should test just one thing and it seems that in this case is testing several scenarios into a single unit test.

Please, split this into 8 or more unit tests.

tests/unit/test_charm.py
```diff
@@ -107,3 +108,340 @@ def test_install_snap_error(self, mock_install):
...
+        test_data["service_protocol"] = "http"
+        self.harness.charm._write_cloud_config(test_data)
+
+        expected_contents["clouds"][CLOUD_NAME]["verify"] = False
+        expected_contents["clouds"][CLOUD_NAME]["auth"]["auth_url"] = (
+            "http://keystone.test:5000/v3"
+        )
+
+        # Check the second call to yaml.dump had the updated content
+        mock_yaml_dump.assert_called_with(expected_contents)
```
gabrielcocenza, line 289: This should also be splitted into two unit tests. One for verify with false and another for verify true.

If both unit tests has very similar structure (seems that this is the case), you can achieve one unit test doing both scenarios by using ` pytest.mark.parametrize` and have the expected output for each scenario

gabrielcocenza (changes requested):
Please split the unit tests. From what I saw they are doing too much things into a singlle test to cover the scenarios

pyproject.toml
```diff
@@ -5,6 +5,7 @@ source = ["."]
 omit = ["tests/**", "docs/**", "lib/**", "snap/**", "build/**", "setup.py"]
 
 [tool.coverage.report]
+fail_under = 80
```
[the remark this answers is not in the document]
H-M-Quang-Ngo: Coverage is now 100% , commit [2fe31ec](https://github.com/canonical/openstack-exporter-operator/pull/142/commits/2fe31ecabdb4ae232c0cfd56146cc54c7180cf59)

tests/unit/test_charm.py
H-M-Quang-Ngo: Thank you! 
For this one, I modified it into 8 scenarios with` pytest.mark.parametrize ` since making multiple separated utests will create lots of duplicated setup.  [d08ced1](https://github.com/canonical/openstack-exporter-operator/pull/142/commits/d08ced1def76cc16e0a80abb690bb0f455cf9dee)

tests/unit/test_charm.py
H-M-Quang-Ngo: Thank you, change has been made. [d08ced1](https://github.com/canonical/openstack-exporter-operator/pull/142/commits/d08ced1def76cc16e0a80abb690bb0f455cf9dee)

jneo8 (commented):
Nice work. One nit non-blocking comment.

tests/unit/test_charm.py
```diff
@@ -107,3 +108,400 @@ def test_install_snap_error(self, mock_install):
...
+            (
+                {"snap_channel": "latest/stable"},  # Default snap channel
+                [("credentials", "keystone"), ("cos-agent", "grafana-agent")],
+                {
+                    "keystone_data": {"random": "data"},
+                    "upstream_present": False,
+                    "snap_present": True,
+                    "snap_active": True,
+                    "resource_path": "/path/to/snap/resource",
+                },
```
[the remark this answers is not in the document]
H-M-Quang-Ngo: Thank you!
1. I just think make it like that helps things look more concise and avoid adding 4 more parameters to the function signature. But maybe splitting them into 5 separate variables make each parameter more explicit in the function signature and improve clarity on what being tested? The change will look like: 
 from this : 
`@pytest.mark.parametrize("config,relations,mocks,expected_status", ...)`
to this:
`@pytest.mark.parametrize("config,relations,keystone_data,upstream_present,snap_present,snap_active,resource_path,expected_status") `
It's a matter of organisation style so, I'm not sure :sweat_smile: . But I'm happy to make change if required, it'll be quick tho.
2.  You're right! So assume the current dict style, I will rename the 'mocks' to something like 'test_values' or 'service_states', does it sound better?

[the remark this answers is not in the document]
jneo8: Thx.

jneo8 (approved):

tests/unit/test_charm.py
```diff
@@ -107,3 +108,400 @@ def test_install_snap_error(self, mock_install):
...
+        self.harness.charm._on_collect_unit_status(mock_event)
+
+        # In cases app_name in expected message
+        if (
+            isinstance(expected_status, ops.BlockedStatus)
+            and "{app_name}" in expected_status.message
+        ):
+            expected_status = ops.BlockedStatus(
+                expected_status.message.format(app_name=self.harness.charm.app.name)
+            )
```
gabrielcocenza, line 274: I don't understand the need of this if condition. Isn't always `openstack-exporter` the application name? I'm almost sure this comes from the charmcraft.yaml

tests/unit/test_charm.py
```diff
@@ -107,3 +108,400 @@ def test_install_snap_error(self, mock_install):
...
+                    "cacert": str(mock_cacert),
+                }
+            }
+        }
+        mock_config.write_text.assert_called_with("yaml content")
+        mock_yaml_dump.assert_called_with(expected_contents)
+
+    @mock.patch("charm.get_installed_snap_service")
+    @mock.patch("charm.snap_install_or_refresh")
+    def test_get_keystone_data(self, _, mock_get_installed_snap_service):
```
gabrielcocenza, line 350: This unit test also has too much asserts.
As a rule of thumb a good unit test should assert just one thing (of course it's not strict).

Please break or create a parameterize as  you did with the other unit test.

tests/unit/test_charm.py
```diff
@@ -107,3 +108,400 @@ def test_install_snap_error(self, mock_install):
...
+
+        # Scenario 7: Remove first relation, should now get data from second relation
+        self.harness.remove_relation(rel_id)
+        result = self.harness.charm._get_keystone_data()
+        assert result == second_rel_data
+
+    @mock.patch("charm.get_installed_snap_service")
+    @mock.patch("charm.OpenstackExporterOperatorCharm._get_keystone_data")
+    @mock.patch("charm.snap_install_or_refresh")
+    def test_snap_service_with_keystone_data(
```
gabrielcocenza, line 433: This is a good example on how we can split this kind of unit test:

`test_snap_service_with_no_keystone_data` 
`test_snap_service_with_keystone_data` 

Even that it might have an overhead of setting the mocks, it's preferable than doing more than one thing and having multiple asserts.

tests/unit/test_charm.py
```diff
@@ -107,3 +108,400 @@ def test_install_snap_error(self, mock_install):
...
+            mock_fetch.return_value = Path("/path/to/snap/resource")
+
+        result = self.harness.charm.get_resource()
+
+        if result is not None:
+            result = result.as_posix()
+
+        assert result == expected_result
+
+        if not model_error:
```
gabrielcocenza, line 498: Please avoid this kind of logic inside unit tests. It's hard to maintain and understand which scenario will enter in a certain condition.

gabrielcocenza (commented):
I really enjoined the refactor that you did. Please continue and apply the same standard for the other functional tests. 

- Parametrize if scenarios are not changing the behavior of the mocks
- break into separate unit tests otherwise

tests/unit/test_charm.py
samuelallan72: +1 if a parametrize the test introduces the requirement for branching logic, then it probably should be separate tests (not parametrized).

tests/unit/test_charm.py
H-M-Quang-Ngo: Oh, that is because I need to set the output status in the pytest decorator of that function, and there is a message that contains the app name inside. But from there we can't directly use a variable that refers to the application name (or there is a way but I don't know yet :thinking: ), and I don't want to just hard-code it, so I did that way :sweat_smile:

samuelallan72:
Small nit: please update the PR title to something short and descriptive. :)

tests/unit/test_charm.py
H-M-Quang-Ngo: Thank you, @samuelallan72 @gabrielcocenza  
Change has been made in [b118a6a](https://github.com/canonical/openstack-exporter-operator/pull/142/commits/b118a6a67c5678a4749daeb1fcaac31beb96c222)
For this one I break it into 4 small separated tests to make the intent clearer.

tests/unit/test_charm.py
H-M-Quang-Ngo: Thank you. I have splitted it into 2 tests [b118a6a](https://github.com/canonical/openstack-exporter-operator/pull/142/commits/b118a6a67c5678a4749daeb1fcaac31beb96c222)

tests/unit/test_charm.py
H-M-Quang-Ngo: Thank you. For this one, consider the complexity of parameterising, I broke it into multiples separated tests for each specific case, and write some helper functions for common setup to reduce duplication.  [b118a6a](https://github.com/canonical/openstack-exporter-operator/pull/142/commits/b118a6a67c5678a4749daeb1fcaac31beb96c222)

tests/unit/test_charm.py
```diff
@@ -107,3 +108,485 @@ def test_install_snap_error(self, mock_install):
...
+        result = self.harness.charm._get_keystone_data()
+        assert result == second_complete_data
+
+    def _setup_common_mocks(self, mocker):
+        """Set up common mocks for _get_keystone_data tests."""
+        mock_get_installed_snap_service = mocker.patch("charm.get_installed_snap_service")
+        mock_upstream = mocker.MagicMock()
+        mock_upstream.present = True
+        mock_get_installed_snap_service.return_value = mock_upstream
+        mocker.patch("charm.snap_install_or_refresh")
```
gabrielcocenza, line 473: Nice :+1:

tests/unit/test_charm.py
gabrielcocenza: I won't block it, but I think you could hard-coded here to use `openstack-exporter`, The chance of changing the name of the project is almost 0.

gabrielcocenza (approved):
Thank you for doing this.

tests/unit/test_charm.py
H-M-Quang-Ngo: If so, doing as I did would be a bit cumbersome and cause unnecessary confusion. I changed it directly as you said :+1:

gabrielcocenza (approved):
Nice job :+1:

chanchiwai-ray (approved):

PR #145, opened 2025-03-19
Add validation to some config options 

### LINKED ISSUE #49: Validating config options
The charm should do the best effort to valid the config options. For example, it's obviously not correct to set the `port` to a negative number.

H-M-Quang-Ngo:
Closes #49  
- Create validation methods for 'port' and 'cache_ttl' config options
- Add validation check to the handler of config_changed event and set block status

ps: 
- 'cache' isn't much to validate for a boolean value and Juju can automatically reject non-boolean values.
-  for 'snap_channel' it's not practical to catch all invalid cases. Also the invalid setup can be noticed easily in Juju logs
- 'ssl_ca' is more complex to validate properly and it's not the charm's duty to verify this but the OpenStack client instead. The integration test also confirms that an invalid CA certificate will cause expected failures

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -11,7 +11,7 @@ OpenStack deployment.
 import logging
 import os
 from pathlib import Path
-from typing import Any, Optional
+from typing import Any, Callable, Optional
 
 import ops
 import yaml
@@ -152,14 +152,14 @@ class OpenstackExporterOperatorCharm(ops.CharmBase):
         Return an error message if any config option is invalid, otherwise None.
 
         """
-        if error := validate_port(self.model.config["port"]):
-            return error
-
-        if error := validate_cache_ttl(self.model.config["cache_ttl"]):
-            return error
-
-        if error := validate_snap_channel(self.model.config["snap_channel"]):
-            return error
+        validators: list[tuple[Callable, str]] = [
+            (validate_port, "port"),
+            (validate_cache_ttl, "cache_ttl"),
+            (validate_snap_channel, "snap_channel"),
+        ]
+        for validator, config_key in validators:
+            if error := validator(self.model.config[config_key]):
+                return error
 
         # All config options are valid
         return None
@@ -169,9 +169,6 @@ class OpenstackExporterOperatorCharm(ops.CharmBase):
         # If this fails, it's not recoverable.
         # So we don't catch the error, instead letting this become a charm error status.
         # Errored hooks are auto-retried by juju, so the install may work on retry.
-        if error := validate_snap_channel(self.model.config["snap_channel"]):
-            logger.error(f"Invalid snap_channel: {error}")
-            return
         snap_install_or_refresh(self.get_resource(), self.model.config["snap_channel"])
 
     def _configure(self, _: ops.HookEvent) -> None:
@@ -182,7 +179,7 @@ class OpenstackExporterOperatorCharm(ops.CharmBase):
         changed.
         """
         if config_error := self.validate_configs():
-            logger.error(f"Cannot configure charm: {config_error}")
+            logger.error(config_error)
             return
 
         self.install()
@@ -227,10 +224,6 @@ class OpenstackExporterOperatorCharm(ops.CharmBase):
 
     def _on_collect_unit_status(self, event: ops.CollectStatusEvent) -> None:
         """Handle collect unit status event (called after every event)."""
-        if config_error := self.validate_configs():
-            event.add_status(BlockedStatus(f"Invalid configuration: {config_error}"))
-            return
-
         if not self.model.relations.get("credentials"):
             event.add_status(BlockedStatus("Keystone is not related"))
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -19,7 +19,7 @@ from charms.grafana_agent.v0.cos_agent import COSAgentProvider
 from ops.model import ActiveStatus, BlockedStatus, ModelError, WaitingStatus
 
 from service import SNAP_NAME, UPSTREAM_SNAP, get_installed_snap_service, snap_install_or_refresh
-from validate_config import validate_cache_ttl, validate_port, validate_snap_channel
+from validate_config import validate_cache_ttl, validate_port
 
 logger = logging.getLogger(__name__)
 
@@ -155,7 +155,6 @@ class OpenstackExporterOperatorCharm(ops.CharmBase):
         validators: list[tuple[Callable, str]] = [
             (validate_port, "port"),
             (validate_cache_ttl, "cache_ttl"),
-            (validate_snap_channel, "snap_channel"),
         ]
         for validator, config_key in validators:
             if error := validator(self.model.config[config_key]):
--- a/src/charm.py
+++ b/src/charm.py
@@ -224,6 +224,9 @@ class OpenstackExporterOperatorCharm(ops.CharmBase):
 
     def _on_collect_unit_status(self, event: ops.CollectStatusEvent) -> None:
         """Handle collect unit status event (called after every event)."""
+        if config_error := self.validate_configs():
+            event.add_status(BlockedStatus(config_error))
+
         if not self.model.relations.get("credentials"):
             event.add_status(BlockedStatus("Keystone is not related"))
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -19,6 +19,7 @@ from charms.grafana_agent.v0.cos_agent import COSAgentProvider
 from ops.model import ActiveStatus, BlockedStatus, ModelError, WaitingStatus
 
 from service import SNAP_NAME, UPSTREAM_SNAP, get_installed_snap_service, snap_install_or_refresh
+from validate_config import validate_cache_ttl, validate_port, validate_snap_channel
 
 logger = logging.getLogger(__name__)
 
@@ -145,11 +146,32 @@ class OpenstackExporterOperatorCharm(ops.CharmBase):
 
         return snap_path
 
+    def validate_configs(self) -> Optional[str]:
+        """Validate the charm config options.
+
+        Return an error message if any config option is invalid, otherwise None.
+
+        """
+        if error := validate_port(self.model.config["port"]):
+            return error
+
+        if error := validate_cache_ttl(self.model.config["cache_ttl"]):
+            return error
+
+        if error := validate_snap_channel(self.model.config["snap_channel"]):
+            return error
+
+        # All config options are valid
+        return None
+
     def install(self) -> None:
         """Install the necessary resources for the charm."""
         # If this fails, it's not recoverable.
         # So we don't catch the error, instead letting this become a charm error status.
         # Errored hooks are auto-retried by juju, so the install may work on retry.
+        if error := validate_snap_channel(self.model.config["snap_channel"]):
+            logger.error(f"Invalid snap_channel: {error}")
+            return
         snap_install_or_refresh(self.get_resource(), self.model.config["snap_channel"])
 
     def _configure(self, _: ops.HookEvent) -> None:
@@ -159,6 +181,10 @@ class OpenstackExporterOperatorCharm(ops.CharmBase):
         It will install the exporter if not already installed or refresh the channel if the it was
         changed.
         """
+        if config_error := self.validate_configs():
+            logger.error(f"Cannot configure charm: {config_error}")
+            return
+
         self.install()
 
         upstream_snap = get_installed_snap_service(UPSTREAM_SNAP)
@@ -201,6 +227,10 @@ class OpenstackExporterOperatorCharm(ops.CharmBase):
 
     def _on_collect_unit_status(self, event: ops.CollectStatusEvent) -> None:
         """Handle collect unit status event (called after every event)."""
+        if config_error := self.validate_configs():
+            event.add_status(BlockedStatus(f"Invalid configuration: {config_error}"))
+            return
+
         if not self.model.relations.get("credentials"):
             event.add_status(BlockedStatus("Keystone is not related"))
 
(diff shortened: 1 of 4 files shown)


chanchiwai-ray:
ps. you don't need to add `Soleng 34` in the title.

src/charm.py
```diff
@@ -159,6 +178,10 @@ def _configure(self, _: ops.HookEvent) -> None:
         It will install the exporter if not already installed or refresh the channel if the it was
         changed.
         """
+        if config_error := self.validate_configs():
```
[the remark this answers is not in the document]
H-M-Quang-Ngo: Thank you. Change has been made [df405eb](https://github.com/canonical/openstack-exporter-operator/pull/145/commits/df405eb79f6d3798dd2bc6738844172e6d35a1ea)

src/charm.py
```diff
@@ -201,6 +224,10 @@ def _on_upgrade(self, _: ops.UpgradeCharmEvent) -> None:
 
     def _on_collect_unit_status(self, event: ops.CollectStatusEvent) -> None:
         """Handle collect unit status event (called after every event)."""
+        if config_error := self.validate_configs():
+            event.add_status(BlockedStatus(config_error))
+            return
```
chanchiwai-ray, line 229: This `return` might not be necessary,  see docs https://ops.readthedocs.io/en/stable/reference/ops.html#ops.CollectStatusEvent

chanchiwai-ray (changes requested):
Final nit from me

src/validate_config.py
```diff
@@ -0,0 +1,46 @@
...
+    return None
+
+
+def validate_cache_ttl(cache_ttl: str) -> Optional[str]:
+    """Validate cache_ttl configuration.
+
+    Return error message if invalid, None if valid.
+
+    """
+    if not re.match(r"^\d+[smhd]$", cache_ttl):
```
samuelallan72, line 26: Validation for the catch ttl is a bit more complex. See https://github.com/canonical/openstack-exporter-operator/blob/209db80d890a81a560b93cb7607dbccda2ea5da6/charmcraft.yaml#L52-L59

The format of https://pkg.go.dev/time#ParseDuration is quite flexible.

Although we could restrict it further in the charm, now that the charm has a stable release, anything more restrictive than what is allowed by golang's ParseDuration will be a breaking change. Eg. if a user has the charm deployed with a cache ttl of `2m30s`, a charm refresh will send the charm into blocked state.

It may not be feasible to validate the cache ttl value in the charm. It's likely the exporter will crash if it's an invalid ttl (this should be tested) - if so, the charm will already go into blocked status, and the user can check the exporter logs and see a message about the invalid value.

src/validate_config.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def validate_snap_channel(snap_channel: str) -> Optional[str]:
+    """Validate snap_channel configuration.
+
+    Return error message if invalid, None if valid.
+
+    """
+    if snap_channel != "latest/stable":
+        if not re.match(r"^(latest/|[^/]+/)(stable|candidate|beta|edge)$", snap_channel):
```
samuelallan72, line 41: I'm not sure about this logic. Why are nested `if` statements required? And it seems the regex would also accept `foo-bar@#$/stable`?

By the way, I'm not sure if we want to restrict the values here, as this prevents potential future cases of a separate channel (eg. v32/beta) or temporary branch (eg. latest/edge/test-fix-bug1).

What's the behaviour of `snap_install_or_refresh` when the channel isn't found? If that already throws an error on an invalid channel, perhaps we can rely on that for validation?

samuelallan72 (changes requested):
I have some comments about validation for cache-ttl and snap_channel. :)

src/validate_config.py
H-M-Quang-Ngo: You're right. It would be more complex and error-prone trying to match all formats the Go ParseDuration supports. Maybe we can can remove the catch ttl validation entirely and let the exporter validate it at runtime. I'll try to examine the behavior for an invalid case.

src/validate_config.py
H-M-Quang-Ngo: Yes, that's wrong regex. Thanks to point out, I forgot to remove the excessive part after latest :)) And nested if is just to ignore the default channel (but yes, it's redundant anw)
Looking at `snap_install_or_refresh` it will raise an error state for invalid snap: 
```
    except snap.SnapError as e:
        logger.error("failed to install snap: %s", str(e))
        raise e
```
It's highly possible the charm will enter an error state requiring manual intervention, so I'm not sure that's desirable or. And also the error log might be buried in logs rather than displayed clearly in charm status.

Do you think we can simplify the validation, like making it accept multiple strings separated by '/'? Or we can simply remove this check :thinking:

H-M-Quang-Ngo:
@chanchiwai-ray @samuelallan72 
I tried to deploy the charm and realised something:
- Setting invalid value for the `snap_channel`: this could be noticed via running `juju debug-log`. The log warning + error messages are quite clear about this, so I think this check can be omitted or made more simple.
-  Setting invalid values for other configs (`port`,` cache_ttl`) : this, however, produced quite tricky behavior. It did sent the charmed to blocked status, with the message saying the snap is not running (this is under the `not snap_service.is_active()` check in the charm code). Otherwise there were no error or warning logs, so in case an invalid config (except for `snap_channel`) happens, it's really hard to find out. 

So I think it's still better to detect early and block with a related message. But for the `cached_ttl`, as said it's complex. Should I implement a more permissive check, like, simply allowing numbers and [valid units](https://pkg.go.dev/time#ParseDuration:~:text=%22ns%22%2C%20%22us%22%20(or%20%22%C2%B5s%22)%2C%20%22ms%22%2C%20%22s%22%2C%20%22m%22%2C%20%22h%22.)?

samuelallan72:
> But for the cached_ttl, as said it's complex. Should I implement a more permissive check, like, simply allowing numbers and [valid units](https://pkg.go.dev/time#ParseDuration:~:text=%22ns%22%2C%20%22us%22%20(or%20%22%C2%B5s%22)%2C%20%22ms%22%2C%20%22s%22%2C%20%22m%22%2C%20%22h%22.)?

Do you think you could do a PoC for validating it completely? It may not be as complex as I originally thought. ParseDuration also allows negative durations, but the charm should not allow that.

We can definitely validate `port`, as that's simple.

> Setting invalid value for the snap_channel: this could be noticed via running juju debug-log. The log warning + error messages are quite clear about this, so I think this check can be omitted or made more simple.

Agreed. I'd probably lean towards a simpler check, just to catch really obvious copy/paste errors for example.

:+1:

H-M-Quang-Ngo:
> Do you think you could do a PoC for validating it completely? It may not be as complex as I originally thought. ParseDuration also allows negative durations, but the charm should not allow that.
 
Thanks for your comments. So I've just looked at Go time library and find the implementation for the [ParseDuration ](https://cs.opensource.google/go/go/+/master:src/time/format.go;l=1616?q=ParseDuration&ss=go%2Fgo#:~:text=//%20%5B%2D%2B%5D%3F(%5B0%2D9%5D*(%5C.%5B0%2D9%5D*)%3F%5Ba%2Dz%5D%2B)%2B), and their regex is `[-+]?([0-9]*(\.[0-9]*)?[a-z]+)+` . I think I can try to translate their implementation into Python, adding negative value check, and then can also take their [test suits](https://cs.opensource.google/go/go/+/master:src/time/time_test.go;l=994?q=ParseDuration&ss=go%2Fgo#:~:text=var%20parseDurationTests%20%3D%20%5B%5Dstruct%20%7B) for validation. How does it sound :thinking:

samuelallan72:
@H-M-Quang-Ngo 

> Thanks for your comments. So I've just looked at Go time library and find the implementation for the [ParseDuration ](https://cs.opensource.google/go/go/+/master:src/time/format.go;l=1616?q=ParseDuration&ss=go%2Fgo#:~:text=//%20%5B%2D%2B%5D%3F(%5B0%2D9%5D*(%5C.%5B0%2D9%5D*)%3F%5Ba%2Dz%5D%2B)%2B), and their regex is [-+]?([0-9]*(\.[0-9]*)?[a-z]+)+ . I think I can try to translate their implementation into Python, adding negative value check, and then can also take their [test suits](https://cs.opensource.google/go/go/+/master:src/time/time_test.go;l=994?q=ParseDuration&ss=go%2Fgo#:~:text=var%20parseDurationTests%20%3D%20%5B%5Dstruct%20%7B) for validation. How does it sound 🤔

Sounds good to me :+1:

chanchiwai-ray (dismissed):
LGTM

src/validate_config.py
```diff
@@ -0,0 +1,77 @@
...
+    Allow patterns in https://pkg.go.dev/time#ParseDuration,
+    Add constraints for negative and zero values.
+    No overflow checks (ParseDuration can handle extremely large values appropriately at runtime)
+
+    Return error message if invalid, None if valid.
+
+    """
+    if not cache_ttl or cache_ttl[0] == "-":
+        return f"Cache_ttl must be non-empty and non-negative. Got {cache_ttl}"
+
```
samuelallan72, line 41: nit: this could be more readable by splitting into two checks, one for non-empty, one for positive.

src/validate_config.py
```diff
@@ -0,0 +1,77 @@
...
+    # Validate overall format
+    if not re.fullmatch(DURATION_PATTERN, cache_ttl):
+        return f"Invalid regex format. Valid pattern is {DURATION_PATTERN}. Got {cache_ttl}"
```
samuelallan72, line 44: I don't think it's very user-friendly to print the regex, as it's long and complex. Perhaps  something along the lines of "cache_ttl is not a valid format. It must be a valid format for https://pkg.go.dev/time#ParseDuration; for example '20m' or '2h30m'."

src/validate_config.py
```diff
@@ -0,0 +1,77 @@
+"""Configuration validation functions."""
+
+import re
+from typing import Optional
+
+MAX_PORT = 65535
+
+# Allowable duration units for cache_ttl from https://pkg.go.dev/time#ParseDuration
+VALID_UNITS = {"ns", "us", "\u05bcs", "\u03bcs", "ms", "s", "m", "h"}
```
samuelallan72, line 9: Where do the extra unicode values come from? The ParseDuration docs say:

> Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h". 

Also, python source files should be valid unicode, so you can directly enter the `µs` value here, and in DURATION_PATTERN. ;)

src/validate_config.py
```diff
@@ -0,0 +1,77 @@
...
+    matches = re.findall(r"(\d+\.?\d*|\d*\.\d+)([a-zµ\u03bc\u05bc]+)", cache_ttl)
+
+    # Validate non-zero duration
+    if not any(float(number) for number, _ in matches):
+        return f"Cache_ttl must be non-zero. Got {cache_ttl}"
+
+    # Validate units
+    for _, unit in matches:
+        if unit not in VALID_UNITS:
+            return f"Invalid time unit {unit}. Valid units are {VALID_UNITS}."
```
samuelallan72, line 56: These messages end up in the juju status output with no extra context; be sure to mention that this relates to the cache_ttl config option.

src/validate_config.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def validate_snap_channel(snap_channel: str) -> Optional[str]:
+    """Validate snap_channel configuration.
+
+    Return error message if invalid, None if valid.
+
+    """
+    if "/" not in snap_channel:
+        return f"Invalid snap_channel format: {snap_channel}. Expected format like 'track/risk'."
```
samuelallan72, line 68: It's actually valid to leave out the risk - you can just specify the track, and you'll get the default risk:

```
sudo snap install charmed-openstack-exporter --channel latest
```

samuelallan72: I'm not sure validation for the snap channel is actually worth it in the end, because there aren't many invalid cases we can actually catch this way.

tests/unit/test_charm.py
```diff
@@ -107,3 +107,112 @@ def test_install_snap_error(self, mock_install):
...
+        """Test config change with valid config."""
+        mocked_upstream_service = mock.MagicMock()
+        mocked_upstream_service.present = True
+        mock_get_installed_snap_service = mocker.patch("charm.get_installed_snap_service")
+        mock_get_installed_snap_service.return_value = mocked_upstream_service
+        mock_install = mocker.patch("charm.OpenstackExporterOperatorCharm.install")
+        mock_event = mock.MagicMock()
+
+        self.harness.begin()
+        self.harness.update_config({config_option: cofig_value})
```
samuelallan72, line 155:
```suggestion
        self.harness.update_config({config_option: config_value})
```

samuelallan72 (changes requested):

src/validate_config.py
H-M-Quang-Ngo: Ah, that's because they mention it explicitly in their [test suite](https://cs.opensource.google/go/go/+/master:src/time/time_test.go;l=994?q=ParseDuration&sq=&ss=go%2Fgo#:~:text=%7B%2212%C2%B5s%22,//%20U%2B03BC). Actually I also used `µs` directly at first, but it failed on the test cases. That's a funny thing, because when I copied their test, these `µ`s look exactly the same, but they turn out to be 2 different unicode values indeed :sweat_smile:

src/validate_config.py
H-M-Quang-Ngo: Oh, I think that's actually a very good one. Should I just copy that whole message :smiley:

src/validate_config.py
H-M-Quang-Ngo: Yes, as said the invalid channel setup can be noticed quite easily in logs. So I think we can indeed remove the snap channel check if there're too many invalid variations to handle.

src/validate_config.py
samuelallan72: Huh interesting :+1:

samuelallan72 (approved):
Nice work! Looks good to me. :)

jneo8 (approved):

Issue #164, opened 2025-09-19
Can't read "Allocated" and "Total" in light mode


WizardBit:
### Description

When inspecting the default "OpenStack Compute Overview" dashboard in light mode it is not easy to read the values in the "Allocated" and "Total" panels. There could possibly be other areas, but these are the two I noticed.

<img width="1900" height="336" alt="Image" src="https://github.com/user-attachments/assets/35e347b8-7c63-4ded-be7c-1cc1ac07220a" />

Inspecting the panel its possible to change the values, but here is the problematic default.

```
        "steps": [
          {
            "color": "rgb(251, 252, 250)",
            "value": null
          }
```

### Charm revision

Revision 58 on latest/stable

### Juju version

3.5.7

### Cloud

OpenStack

### Expected behaviour

Visible text that is easy to read

### Reproduce / Test

Deploy the openstack-exporter charm and change to the light mode view for grafana

PR #166, opened 2025-10-31
fix: panel colors for `cloud` and `compute` dashboards

Deezzir:
Closes: #164


Pjack (commented):
I remember we need to patch sunbeam [src](https://opendev.org/openstack/sunbeam-charms/src/branch/main/charms/openstack-exporter-k8s/src) first and sync to this project by terraform. 

For example
* https://github.com/canonical/openstack-exporter-operator/pull/161

PR #189, opened 2026-03-10
feat: add Octavia LoadBalancer alert rules for provisioning errors and pending states

marcusboden:
We've tested these on PS6 for quite a while and they seem to work quite well.


gabrielcocenza:
Closing to use the automation from sunbeam in: #190

PR #191, opened 2026-03-16
feat: add SSDLC system event logging

Deezzir:
Implements SSDLC (Secure Software Development Lifecycle) logging to track critical system events including startup, shutdown, restart, and crash events for the `charmed-openstack-exporter` snap

### DIFF
--- a/src/service.py
+++ b/src/service.py
@@ -6,6 +6,8 @@ from typing import Any, Optional
 
 from charms.operator_libs_linux.v2 import snap
 
+from ssdlc import SSDLCSysEvent, log_ssdlc_system_event
+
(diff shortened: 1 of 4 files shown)


gabrielcocenza (dismissed):

H-M-Quang-Ngo (dismissed):

gabrielcocenza (approved):
There is a steps that I think might worth adding some logs:

- when the charm writed the keystone config for authenticate `_write_cloud_config` maybe this is kind of Token Created situation?


Otherwise LGTM

H-M-Quang-Ngo (approved):

Issue #192, opened 2026-03-31
openstack service dashboard not reporting Agent status


mathmarchand:
### Description

Dashboard is missing data for the agent status. The metrics seems to have changed name from openstack_nova_agent to openstack_nova_agent_total, simlarly for neutron and cinder.

### Charm revision

latest/stable (rev. 59)

### Juju version

3.6.20

### Cloud

OpenStack

### Expected behaviour

Dashboard display the proper value for the agent status.

### Reproduce / Test

Install openstack exporter on Charmed openstack Caracal with openstack-service-checks revision 81 and relate it to cos/grafana:grafana-dashboard. 

The metrics have change names, but the dashboard still pointing to the old one.

Deezzir:
@mathmarchand https://github.com/canonical/opentelemetry-collector-k8s-operator/issues/232 should've fixed the issue, can you confirm?

mathmarchand:
@Deezzir yes, the customer upgraded to the latest release and it's fixed. 

Thank you!

PR #193, opened 2026-03-31
Update service.json for new agent metrics names

mathmarchand:
Nova, Cinder and Neutron agent metric names changed to openstack_<service>_agent_total from openstack_<service>_agent.

### DIFF
--- a/src/grafana_dashboards/service.json
+++ b/src/grafana_dashboards/service.json
@@ -94,7 +94,7 @@
       "tableColumn": "",
       "targets": [
         {
-          "expr": "count(openstack_nova_agent_state{adminState=\"enabled\"})-sum(openstack_nova_agent_state{adminState=\"enabled\"})",
+          "expr": "count(openstack_nova_agent_state_total{adminState=\"enabled\"})-sum(openstack_nova_agent_state_total{adminState=\"enabled\"})",
           "format": "time_series",
           "hide": false,
           "instant": true,
@@ -236,7 +236,7 @@
       ],
       "targets": [
         {
-          "expr": "openstack_nova_agent_state{adminState=\"enabled\"}",
+          "expr": "openstack_nova_agent_state_total{adminState=\"enabled\"}",
           "format": "table",
           "instant": true,
           "intervalFactor": 1,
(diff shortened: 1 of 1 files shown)


mathmarchand:
Related to this [issue](https://github.com/canonical/openstack-exporter-operator/issues/192) .

gabrielcocenza (changes requested):
Thanks for this patch @mathmarchand. May I ask you some things?
1 - Contribute to the upstream  to fix the [Readme](https://github.com/openstack-exporter/openstack-exporter/?tab=readme-ov-file#metrics-collected)
2 - Share a print in the PR with the prometheus metrics or the dashboard properly 
3 - Update the Prometheus alert rules in [sunbeam](https://opendev.org/openstack/sunbeam-charms/src/branch/main/charms/openstack-exporter-k8s/src/prometheus_alert_rules). (We sync the alert rules from there)

mathmarchand:
Hi @gabrielcocenza,

Some more context, it seems that addition of _total is not from upstream openstack-exporter code.:

The Prometheus Go client library enforces the OpenMetrics specification, which requires that all counter-type metrics must end in _total. So when the metric type is declared as CounterValue, the Go client automatically appends _total to the metric name at scrape time.

This means:

The metric is defined in code as agent_state
The exporter registers it as openstack_nova_agent_state
Prometheus scrapes it as openstack_nova_agent_state_total

Basically if prometheus is on version 2.x+, the _total metric need to be used, if not the old metric need to be used. Since, the COS Lite 1/stable is EOL and we should be using 2/stable until the new LTS version is released, this _metrics is needed.

I will update the query so that it could work in both situations.

mathmarchand:
As requested, here's a output of the adjusted query in Grafana.
Screenshot from 2026-04-01 09-23-33

Deezzir:
@mathmarchand The addition of `_total` suffix is coming from `opentelemetry-collector.` We pushed the change to the collector to disable the normalization for now. Please see https://github.com/canonical/opentelemetry-collector-k8s-operator/pull/238 and https://github.com/canonical/opentelemetry-collector-k8s-operator/issues/232. The change you're proposing shouldn't be needed anymore.

P.S. If we merge the PR, the backward compatibility with `grafana-agent` will be broken.

PR #203, opened 2026-06-30
test: adopt -Werror in unit tests

tonyandrewmeyer:
Add [tool.pytest.ini_options] filterwarnings = ["error", ...] so the unit suite fails on deprecations and resource leaks, with two narrow, message-anchored ignores:

- Harness is deprecated (PendingDeprecationWarning): the intentional migration signal; ignored until tests/unit/test_charm.py moves to Scenario.
- GrafanaDashboard._serialize is deprecated (DeprecationWarning): the vendored grafana_agent/v0/cos_agent.py lib calls cosl's deprecated serializer; fix belongs upstream in grafana-agent-operator / cosl.

This bumps ops from 2.x to 3.x -- it seems like that is ok, the tests don't fail and it doesn't conflict with a minimum Python version. If that can't be done, then this change can be done with ops 2.x with one other ignore (or a new version of 2.23 at some point).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ version = "0.0.0"
 requires-python = ">=3.10"
 dependencies = [
     "cosl",
-    "ops ~= 3.8",
+    "ops ~= 2.5",
     "pydantic < 2",
     "pyyaml ~= 6.0",
 ]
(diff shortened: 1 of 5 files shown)


jneo8:
The functional test failed because of zaza dependency issue.

This remind me may be it's a good timing for us to migrate to jubilant: https://github.com/canonical/openstack-exporter-operator/issues/206

